PR #10, opened 2026-03-17
feat: add inference-api-translator BBR plugin with Anthropic support

noyitz:
Implement a BBR plugin that translates inference API requests and
  responses between OpenAI Chat Completions format and provider-native
  formats. The plugin reads the X-Gateway-Destination-Provider header
  to determine if translation is needed, and auto-detects the provider
  format on the response path.

  Includes Anthropic Messages API translation (request and response),
  OpenAI passthrough provider, and an extensible Provider interface
  for adding future providers (Bedrock, Gemini).

## Merge criteria:

## Release Notes

* **New Features**
  * Added inference API translator plugin supporting OpenAI and Anthropic provider formats, enabling automatic API request and response translation.

* **Tests**
  * Added comprehensive test suite covering translation scenarios, provider integration, and error handling.

* **Chores**
  * Added Docker containerization support and updated project dependencies.


pkg/plugins/inference_api_translator/plugin.go
```diff
@@ -0,0 +1,179 @@
+/*
+Copyright 2026 The opendatahub.io Authors.
```
nirrozenbaum: rename dir and functions accordingly (api-translation for dir, api_translation for package alias)

pkg/plugins/inference_api_translator/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	PluginType = "inference-api-translator"
+
+	// ProviderHeader is the request header that indicates the target provider.
+	// Set by an upstream plugin in the BBR chain (e.g., api-key-injection).
+	// Absent or "openai" means no translation needed.
+	ProviderHeader = "X-Gateway-Destination-Provider"
+)
+
+// compile-time type validation
+var _ framework.RequestProcessor = &InferenceAPITranslatorPlugin{}
```
nirrozenbaum: APITranslationPlugin

pkg/plugins/inference_api_translator/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+
+	translatedBody, err := provider.TranslateResponse(response.Body, model)
+	if err != nil {
+		return fmt.Errorf("response translation failed for provider %q: %w", providerName, err)
+	}
+
+	if translatedBody != nil {
+		response.SetBody(translatedBody)
+	}
+
```
nirrozenbaum: ditto - what about headers?

pkg/plugins/api_translation/tests/plugin_test.go
```diff
@@ -0,0 +1,283 @@
...
+	assert.Equal(t, "application/json", mutated["content-type"])
+
+	removed := req.RemovedHeaders()
+	assert.Contains(t, removed, "authorization")
+	assert.Contains(t, removed, "content-length")
+
+	// Verify model stored in CycleState
+	model, _ := framework.ReadCycleStateKey[string](cs, "api-translation-model")
+	assert.Equal(t, "claude-sonnet-4-20250514", model)
+}
```
nirrozenbaum: is this stored by the api-translation plugin?
we should let the model resolver plugin later to write to CycleState. 
for now it's fine to unblock the functionality on the response translation.

pkg/plugins/api_translation/tests/plugin_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+func TestProcessRequest_UnknownProvider(t *testing.T) {
+	p, err := plugin.NewAPITranslationPlugin()
+	require.NoError(t, err)
+
+	cs := newCycleStateWithProvider("bedrock")
```
nirrozenbaum: can we use explicitly something like:

```suggestion
	cs := newCycleStateWithProvider("unknown")
```

we're going to have soon bedrock which will break this test

pkg/plugins/api_translation/tests/plugin_test.go
```diff
@@ -0,0 +1,283 @@
...
+	errObj := resp.Body["error"].(map[string]any)
+	assert.Equal(t, "invalid_request_error", errObj["type"])
+}
+
+func TestProcessResponse_AnthropicToolUse(t *testing.T) {
+	p, err := plugin.NewAPITranslationPlugin()
+	require.NoError(t, err)
+
+	cs := newCycleStateWithProvider("anthropic")
+	cs.Write("api-translation-model", "claude-sonnet-4-20250514")
```
nirrozenbaum: why do you write here to cycle state and in the other test you dont?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY cmd/ cmd/
+COPY pkg/ pkg/
+
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
+    -ldflags="-w -s" \
+    -o /bbr-server \
+    ./cmd/main.go
+
+FROM ${BASE_IMAGE}
```
nirrozenbaum: I want to do minor changes to the Dockerfile. 
don't want to block the PR on that, but giving a heads up.
for example, the build should be inside a container and not on your mac. so you shouldn't have bbr-server locally on your mac after building the image. 
and then we also don't need these lines on .gitignore. 

generally speaking we should use multi stage docker build and avoid any local builds cause each laptop may have different hw.

pkg/plugins/api_translation/plugin.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/api_translation/providers/anthropic"
+)
+
+const (
+	// PluginType is the unique identifier for this plugin, used in CLI flags and registry.
+	PluginType = "api-translation"
+
+	// CycleState keys
+	cycleStateProviderKey = "provider"             // read from CycleState, set by upstream plugin
+	cycleStateModelKey    = "api-translation-model" // written by this plugin for response use
```
nirrozenbaum: nit: why not call it jus `model`?
this would be used by another plugins as well (like key injection).

PR #13, opened 2026-03-18
feat(apikey): add api-key-injection BBR plugin

### LINKED ISSUE #4: implement api-key injection as payload processing request plugin

yehuditkerido:
Inject provider-specific API key headers into inference requests based on annotated Kubernetes Secrets. Updates example_plugin signatures to match current upstream framework API.

## Description

Add a new `api-key-injection` BBR plugin (`pkg/plugins/apikey/`) that watches annotated Kubernetes Secrets and injects provider-specific API key headers into inference requests at runtime. 

Also updates `example_plugin.go` signatures and pins `gateway-api-inference-extension` to commit `5d9d3543` to align with the current upstream framework API.

## How Has This Been Tested?

- Unit tests: `go test ./pkg/plugins/apikey/...` — all pass
- Build: `go build ./...` — clean
- Local Kind cluster: verified Secret reconciler picks up annotated Secrets correctly

Resolves #4 

## Summary by CodeRabbit

* **New Features**
  * Enabled API key injection plugin for application use.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/bbr/framework"
 
 	api_translation "github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/api-translation"
+	apikey_injection "github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/apikey_injection"
 	provider_resolver "github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/provider-resolver"
 )
 
@@ -45,7 +46,7 @@ func main() {
 }
 
 func registerPlugins() {
-	// framework.Register(plugins.ExamplePluginType, plugins.ExamplePluginFactory) // example plugin, can be removed later
 	framework.Register(provider_resolver.ProviderResolverPluginType, provider_resolver.ProviderResolverFactory)
 	framework.Register(api_translation.APITranslationPluginType, api_translation.APITranslationFactory)
+	framework.Register(apikey_injection.APIKeyInjectionPluginType, apikey_injection.APIKeyInjectionFactory)
 }
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.25.0
 
 require (
 	github.com/stretchr/testify v1.11.1
+	k8s.io/api v0.35.2
 	k8s.io/apimachinery v0.35.2
 	sigs.k8s.io/controller-runtime v0.23.3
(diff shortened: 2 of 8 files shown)


noyitz:
lets hold the merge until we resolve those two issues:
- Duplicated data between MaaSModelRef and Secret
- Provider should come from CycleState (and follow the other plugins approach), not from Secret annotations

The plugin chain should be:
  1. provider-resolver — reads MaaSModelRef, writes provider, credentialRef (name + namespace) to CycleState
  2. api-translation — reads provider from CycleState, translates body/headers if needed
  3. api-key-injection — reads provider and credentialRef from CycleState, fetches the Secret by the exact name/namespace from credentialRef, reads data["api-key"], injects the right header format
more info was sent separately

nirrozenbaum:
need to rebase :/

pkg/plugins/apikey/injector.go
```diff
@@ -0,0 +1,66 @@
...
+const (
+	// Bearer-based provider: "Authorization: Bearer <key>"
+	ProviderOpenAI = "openai"
+
+	// Raw-header provider: key goes in a provider-specific header
+	ProviderAnthropic = "anthropic" // x-api-key: <key>
+
+	// DefaultProvider is used when a Secret has no provider annotation.
+	DefaultProvider = ProviderOpenAI
+)
```
nirrozenbaum: I've moved provider consts into a dedicated package that could be referenced by multiple plugins (which also ensures consistency between plugins).
can you delete these consts and instead use this?
https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/1c162bcaac3fca4463852a52b45c0d5f4fc50ad1/pkg/external-model/provider/provider.go#L3-L6

pkg/plugins/apikey/injector.go
```diff
@@ -0,0 +1,55 @@
...
+// RawHeaderInjector sets one or more configurable headers.
+// Used by providers that expect the key in a non-standard header
+// (e.g. Anthropic uses "x-api-key").
+// ExtraHeaders allows injecting additional provider-specific headers.
+type RawHeaderInjector struct {
+	HeaderName   string
+	ExtraHeaders map[string]string
+}
+
+func (r *RawHeaderInjector) Inject(apiKey string) map[string]string {
```
nirrozenbaum: I'm starting to have doubts on the extra headers - although I was the one who asked it 🫤.
after seeing things come together more and more, the apiversion seems to be in the scope of api translation. not the api-key injection.

at this point, I would recommend we have a single injector with HeaderName and HeaderValuePrefix.
then we can use a single injector.

apologize 🙏

PR #34, opened 2026-03-24
[Feat]: add Azure OpenAI E2E test with llm-katan simulator

abdallahsamabd:
## Description

Adds an end-to-end test for the Azure OpenAI api-translation provider, validating the full request flow through the gateway infrastructure:

**Client → Envoy → BBR ext_proc → llm-katan (simulated Azure OpenAI backend)**

The test uses [llm-katan](https://github.com/opendatahub-io/llm-katan) in echo mode as a Docker container to simulate an Azure OpenAI endpoint. No external endpoints, real API keys, or Kubernetes cluster are required.

**What the test validates:**
- Path rewriting from `/v1/chat/completions` (OpenAI) to `/openai/deployments/{model}/chat/completions?api-version=2024-10-21` (Azure)
- API key header injection (`api-key` header for Azure instead of `Authorization: Bearer`)
- Request body passthrough (Azure uses the same schema as OpenAI — no body mutation)
- Response passthrough (Azure returns OpenAI-format responses — no translation needed)
- Full plugin chain execution: `ProviderSetter → ApiKeySetter → ApiTranslation`

**Components:**
| Component | How it runs |
|---|---|
| HTTP client | In-process (Go test) |
| Envoy proxy | Docker container (ext_proc filter) |
| BBR ext_proc server | In-process (gRPC, with production ApiTranslation plugin) |
| llm-katan | Docker container (echo mode, no real model) |

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test validating the Azure OpenAI request/response flow through the stack using Docker and Envoy.
  * Added test helpers to set provider and inject API keys to exercise provider-specific authentication in e2e runs.
  * Added an Envoy configuration template and a lightweight test image to support the e2e scenarios.

### DIFF
--- a/test/e2e/azure_e2e_test.go
+++ b/test/e2e/azure_e2e_test.go
@@ -1,0 +1,407 @@
+/*
+Copyright 2026 The opendatahub.io Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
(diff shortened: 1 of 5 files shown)


nirrozenbaum:
@abdallahsamabd we now have e2e tests using llm-m katan that runs on aws instance.
do you think we still need this PR?

abdallahsamabd:
@nirrozenbaum
in this case I think the current PR is not needed and it can be closed
thanks

PR #93, opened 2026-03-31
test(apikey-injection): refactor store tests for clarity

yehuditkerido:
## Description
Refactor `store_test.go` to follow proper table-driven test patterns:
- Separate setup from validation in `TestSecretStore`
- Remove duplicate tests between `TestSecretStore` and `TestAddOrUpdate`
- Create focused `TestAddOrUpdateErrors` for error cases only

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation tightened for API key injection: empty credential fields are now treated as explicit errors, preventing storage of secrets with empty values.

* **Tests**
  * Reworked test suite for the API key injection plugin: moved to a shared assertion pattern, reduced and clarified error-focused cases, and renamed one test to reflect its error-centric coverage.

### DIFF
--- a/pkg/plugins/apikey-injection/store.go
+++ b/pkg/plugins/apikey-injection/store.go
@@ -38,22 +38,22 @@ func newSecretStore() *secretStore {
 	}
 }
 
-// addOrUpdate extracts all non-empty fields from the Secret's data and stores
-// them under the given key. Returns an error if the Secret has no data fields.
+// addOrUpdate extracts all fields from the Secret's data and stores them under
+// the given key. Returns an error if the Secret has no data fields or if any
+// field is empty.
 func (s *secretStore) addOrUpdate(key string, secret *corev1.Secret) error {
 	if len(secret.Data) == 0 {
-		return fmt.Errorf("secret %s has no data fields", key)
+		s.delete(key)
+		return fmt.Errorf("secret '%s' has no data fields", key)
 	}
 
 	credentials := make(map[string]string)
 	for field, value := range secret.Data {
-		if len(value) > 0 {
-			credentials[field] = string(value)
+		if len(value) == 0 {
+			s.delete(key)
+			return fmt.Errorf("secret '%s' has empty field '%s'", key, field)
 		}
-	}
-
-	if len(credentials) == 0 {
-		return fmt.Errorf("secret %s has no non-empty data fields", key)
+		credentials[field] = string(value)
 	}
 
 	s.mu.Lock()
--- a/pkg/plugins/apikey-injection/store_test.go
+++ b/pkg/plugins/apikey-injection/store_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
 
 func TestSecretStore(t *testing.T) {
 	tests := []struct {
-		name          string
-		initStoreFunc func(t *testing.T, s *secretStore)
+		name      string
+		setup     func(s *secretStore)
```
nirrozenbaum: previous name was clearer :) 
(`initStoreFunc`)

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
+		getKey    string
```
nirrozenbaum: `secretKey`?

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
+			wantCreds: map[string]string{"api-key": "sk-key-1"},
 		},
 		{
-			name: "get nonexistent key returns not found",
-			initStoreFunc: func(t *testing.T, s *secretStore) {
-				_, found := s.get("default/nonexistent")
-				assert.False(t, found)
-			},
+			name:      "get nonexistent key returns not found",
+			setup:     nil,
```
nirrozenbaum: in tests it's better to have empty func and avoid nil checks which is less readable.
e.g.,
```go
setup: func(s *secretStore) {}
```

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
-				assert.True(t, f2)
-				assert.Equal(t, "value-1", v1["api-key"])
-				assert.Equal(t, "value-2", v2["api-key"])
-
-				s.delete("default/key-a")
-				_, f1 = s.get("default/key-a")
-				_, f2 = s.get("default/key-b")
-				assert.False(t, f1)
-				assert.True(t, f2)
-			},
```
nirrozenbaum, line 121: that was a nice test. 
it doesn't fit in this test setup, but maybe we can create another Test function and put it back?

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			s := newSecretStore()
-			tt.initStoreFunc(t, s)
+			if tt.setup != nil {
+				tt.setup(s)
+			}
```
nirrozenbaum: revert after you remove the nil entry.
e.g., 
```go
tt.initStoreFunc(s) 
```

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
+
+			creds, found := s.get(tt.getKey)
+			assert.Equal(t, tt.wantFound, found)
+			if tt.wantFound && tt.wantCreds != nil {
```
nirrozenbaum: if wantFound should be enough if the test is set up correctly.
we're not testing the test setup itself :)  
```suggestion
			if tt.wantFound {
```

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,135 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
-				Data: map[string][]byte{},
+				ObjectMeta: metav1.ObjectMeta{Name: "no-data", Namespace: "default"},
+				Data:       map[string][]byte{},
 			},
-			wantKey: "default/no-data",
-			wantErr: true,
+			key: "default/no-data",
 		},
 		{
 			name: "returns error when all fields are empty",
```
nirrozenbaum: I was reading store.go again and I think we have some minor issue here.
if a secret contains multiple fields and some are empty, we have a silent failure. 
we probably want to fail if any field is empty (secret field with empty string shouldn't be valid even if there's a different valid field).

nirrozenbaum (commented):
@yehuditkerido reviewed the PR, overall looks good. 
most of the comments are about style, but one comment is about functionality. I think we should fail addOrUpdate when any field contains empty string. 
otherwise we fail silently.

pkg/plugins/apikey-injection/store_test.go
yehuditkerido: Actually I  thought it was a redundant test since eventually it test go maps behavior, correct me if I'm wrong. It suppose to test independency of secrets that were build based on go maps. Isn't that too trivial? 
Happy to restore it if you disagree.

pkg/plugins/apikey-injection/store_test.go
yehuditkerido: Great catch. Fixed - now fails early if any field is empty.

pkg/plugins/apikey-injection/store_test.go
nirrozenbaum: mmm. we could also argue that the tests should not care about internal implementation details. so in theory we could change from map to a different storage and the test should remain the same.
but given the simplicity of the store, that's an acceptable argument. so we can keep as is without it.

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,132 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
 			s := newSecretStore()
-			tt.initStoreFunc(t, s)
+		tt.initStoreFunc(s)
+
+			creds, found := s.get(tt.secretKey)
+			assert.Equal(t, tt.wantFound, found)
+		if tt.wantFound {
+			if diff := cmp.Diff(tt.wantCreds, creds); diff != "" {
+				t.Errorf("credentials mismatch (-want +got):\n%s", diff)
+			}
```
nirrozenbaum: we need to add here cmpopts sort otherwise we might have flakes in the tests...
this function compares maps including their order (sort verifies the order is identical in both maps)

pkg/plugins/apikey-injection/store_test.go
```diff
@@ -50,193 +50,132 @@ func testSecret(namespace, name string, credentials map[string]string) *corev1.S
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			s := newSecretStore()
-			tt.initStoreFunc(t, s)
+		tt.initStoreFunc(s)
+
+			creds, found := s.get(tt.secretKey)
+			assert.Equal(t, tt.wantFound, found)
```
nirrozenbaum: nit: `require.Equal`

pkg/plugins/apikey-injection/store.go
```diff
@@ -38,22 +38,20 @@ func newSecretStore() *secretStore {
...
 }
 
-// addOrUpdate extracts all non-empty fields from the Secret's data and stores
-// them under the given key. Returns an error if the Secret has no data fields.
+// addOrUpdate extracts all fields from the Secret's data and stores them under
+// the given key. Returns an error if the Secret has no data fields or if any
+// field is empty.
 func (s *secretStore) addOrUpdate(key string, secret *corev1.Secret) error {
 	if len(secret.Data) == 0 {
 		return fmt.Errorf("secret %s has no data fields", key)
```
nirrozenbaum: super nit:
```suggestion
		return fmt.Errorf("secret '%s' has no data fields", key)
```

pkg/plugins/apikey-injection/store.go
```diff
@@ -38,22 +38,20 @@ func newSecretStore() *secretStore {
...
 	}
 
 	credentials := make(map[string]string)
 	for field, value := range secret.Data {
-		if len(value) > 0 {
-			credentials[field] = string(value)
+		if len(value) == 0 {
+			return fmt.Errorf("secret %s has empty field %s", key, field)
```
nirrozenbaum: super nit:
```suggestion
			return fmt.Errorf("secret '%s' has empty field '%s'", key, field)
```

nirrozenbaum (commented):
one real comment about possible test flaks when the secret contains multiple fields. 
I wouldn't add the nits if it wasn't for that comment.. but already added a comment so why not :)

nirrozenbaum (approved):
nice :)

/lgtm
/approve

nirrozenbaum:
@noyitz can you please have a look why e2e fail?
it looks like timeout. maybe something is missing in the configuration?

nirrozenbaum:
overriding the failing e2e test which is not related to this PR and was fixed in #124

PR #111, opened 2026-04-07
refactor model resolver plugin

### LINKED ISSUE #110: ambiguity problem when multiple ExternalModel CRs are using the same name
ExternalModel CR is a namespaced scoped resource. 
therefore, we could have different external models having the same name in different namespaces.
when a request comes in to payload processing, model resolver plugin extracts the model information based on the model from the request body. this ignores the origin `MaaSModelRef`. 

Current approach may introduce ambiguity: 
when an incoming request comes in with `model: gpt`, this model could be defined in multiple namespaces. 
we must update the code to take into consideration the original path of the request, parse from it the `MaaSModelRef` namespaced name and select the correct ExternalModel.

This problem originates from MaaS. model name is not forced to be unique in maas, but rather the url path includes the namespaced name of the `MaaSModelRef`. 
as part of the bigger solution in MaaS, we should remove the option to use same model identifier more than once (in different namespaces). I would even consider making `MaaSModelRef` a cluster wide resource to guarantee uniqueness.

until that happens, we should add a reconciler to watch `MaaSModelRef` in payload processing code, and match to the correct ExternalModel based on the relative path of the request. we should also make the adaptations to the model info store.

cc: @noyitz @jland-redhat

nirrozenbaum:
## Description
current code make implicit assumptions that are wrong, like the fact that MaaSModelRef name is always identical to ExternalModel name. 
also, when getting an incoming request, the existing code checks in model store without any distinction of which MaaSModelRef the request refers to. 
this is buggy, since two MaaSModelRef objects in different namesapaces may use the same name and may refer to the same model name. 
this PR fixes these issues. 

fixes: #110 

## Merge criteria:


nirrozenbaum:
on hold until https://github.com/opendatahub-io/models-as-a-service/pull/709 is merged

szedan-rh (approved):
LGTM

Issue #114, opened 2026-04-10
external model simulator: add the ability for dummy api key validation


nirrozenbaum:
the current external model simulator simulates multiple external providers which is great.
in terms of api key - 
current sim only checks the header is there but no validation is done on the value during the request handling.
to have a "closer to reality" simulator, we should add the option to configure what is the expected api key in the sim, per provider.
this should allow us to test in e2e the actual api key injection (that the right key was injected) and not only presence of the right header.

cc: @yossiovadia @noyitz

yossiovadia:
I'll take it

yossiovadia:
once #115 will be merged, i'll replace the AWS instance with the new simulator that enforces usage of correct key.

nirrozenbaum:
closed as done in #115

PR #115, opened 2026-04-10
Align e2e test keys with llm-katan simulator defaults

yossiovadia:
## Summary

- Update e2e test API key secrets to match llm-katan simulator default keys
- Update Azure live mock test to use the correct default key
- Add invalid key rejection test
- Document `--validate-keys`, `--api-keys`, and `--tls` in simulator README

## Context

The llm-katan simulator ([v0.10.0](https://pypi.org/project/llm-katan/0.10.0/)) now supports API key **value** validation via `--validate-keys` ([GitHub](https://github.com/yossiovadia/llm-katan)). Previously it only checked that auth headers were present — now it can validate the actual key value matches what's expected.

When a wrong key is sent, the error response includes the expected key:
```json
{"error": {"message": "invalid API key for openai: got 'wrong-key', expected 'llm-katan-openai-key'"}}
```

This is safe for a test simulator (not a security boundary) and makes debugging e2e failures trivial.

### How it works

```bash
# Without --validate-keys (current behavior, unchanged)
llm-katan --model test --backend echo --providers openai,anthropic
# Any key value accepted, only header presence checked

# With --validate-keys (opt-in)
llm-katan --model test --backend echo --providers openai,anthropic --validate-keys
# Key values validated against defaults, wrong keys get 401 with hint
```

Default keys per provider:

| Provider | Default Key |
|----------|------------|
| OpenAI | `llm-katan-openai-key` |
| Anthropic | `llm-katan-anthropic-key` |
| Vertex AI | `llm-katan-vertexai-key` |
| Bedrock | `llm-katan-bedrock-key` |
| Azure OpenAI | `llm-katan-azure-key` |

## Changes

| File | Change |
|------|--------|
| `test/e2e/e2e_test.go` | Secrets use `llm-katan-{provider}-key` instead of `e2e-test-key-*` |
| `test/e2e/e2e_test.go` | New test: sends wrong key, expects 401 with key hint |
| `pkg/.../azure/azure_openai_test.go` | `test-key` → `llm-katan-azure-key` |
| `tools/simulator/README.md` | Document `--validate-keys`, `--api-keys`, `--tls`, default key table |

## Backward compatible

- Without `--validate-keys`, simulator accepts any key value (current behavior)
- E2e tests work in both modes: correct keys pass with validation on, any keys pass with validation off
- The invalid key test handles both: expects 401 when validation is on, 200 when off

Related: #114

## Requires

llm-katan >= 0.10.0: `pip install llm-katan>=0.10.0`

## Summary by CodeRabbit

* **New Features**
  * Added simulator CLI options: --validate-keys, --api-keys, and --tls for HTTPS with self-signed certs.

* **Documentation**
  * Clarified simulator authentication behavior and documented the new CLI options and per-provider default keys.

* **Tests**
  * Updated tests to use simulator default keys and added an end-to-end scenario that validates API key verification when enabled.

### DIFF
--- a/pkg/plugins/api-translation/translator/azure/azure_openai_test.go
+++ b/pkg/plugins/api-translation/translator/azure/azure_openai_test.go
@@ -413,7 +413,7 @@ func TestTranslateResponse_LiveMockIntegration(t *testing.T) {
 	req, err := http.NewRequest("POST", url, bytes.NewReader(reqBody))
 	require.NoError(t, err)
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("api-key", "test-key")
+	req.Header.Set("api-key", "llm-katan-azure-key")
 
 	client := &http.Client{Timeout: 10 * time.Second}
 	resp, err := client.Do(req)
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -13,21 +13,32 @@ import (
 
 // Provider represents a provider configuration for testing.
 type Provider struct {
-	Name     string
-	Provider string
+	Name         string
+	Provider     string
+	SimulatorKey string // expected API key for llm-katan --validate-keys
+}
+
+// simulatorKey maps provider names to llm-katan default keys.
+// These match the DEFAULT_API_KEYS in llm-katan's config.py.
+var simulatorKeys = map[string]string{
+	"openai":       "llm-katan-openai-key",
+	"anthropic":    "llm-katan-anthropic-key",
+	"azure-openai": "llm-katan-azure-key",
+	"vertex":       "llm-katan-vertexai-key",
+	"bedrock-openai": "llm-katan-bedrock-key",
 }
 
 var providers = []Provider{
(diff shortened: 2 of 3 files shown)


noyitz (commented):
Looks good overall — clean changes, well-documented README additions.

Two notes:

**1. Invalid key test is too permissive:**
The assertion `strings.Contains(resp, "401") || strings.Contains(resp, "200")` always passes regardless of whether `--validate-keys` is enabled on the simulator. If the intent is to test key validation, consider either:
- Gating the test on an env var (e.g. `E2E_SIMULATOR_VALIDATE_KEYS=true`) and asserting 401 only
- Or skipping the test when key validation isn't enabled

As written, it documents the behavior but doesn't actually catch regressions.

**2. Minor style nit (non-blocking):** The `simulatorKeys` map is only used to populate the `providers` slice. Could simplify by putting the key values inline directly on each `Provider`. But totally fine as-is — the map makes the key-to-provider mapping easy to reference.

LGTM with the note about the test assertion.

PR #113, opened 2026-04-09
feat(guardrails): add response-nemo-guardrails bbr plugin

liavweiss:
### Summary

- Add a new `nemo-response-guard` BBR plugin that checks model responses against NeMo Guardrails **output rails** before returning them to the caller
- Extract shared HTTP/config logic from `nemo-request-guard` into a common `nemo_guard.go` file to eliminate code duplication between the two plugins
- Add deployment documentation for using both plugins with the TrustyAI Guardrails Operator

### What changed

**New file: `pkg/plugins/nemo/response_guard.go`**

Implements `framework.ResponseProcessor`. After the model backend responds, this plugin:
1. Extracts the assistant message from `choices[0].message.content` in the OpenAI-style response body
2. POSTs it to NeMo's `/v1/guardrail/checks` endpoint with `{"role": "assistant"}` so NeMo routes it to **output** rails
3. If NeMo returns `"status": "success"` the response proceeds; any other status returns HTTP 403

**New file: `pkg/plugins/nemo/nemo_guard.go`** (shared code)

Refactored common code between `nemo-request-guard` and `nemo-response-guard` into a shared base:
- `nemoGuardBase` struct — holds `typedName`, `nemoURL`, `httpClient`
- `nemoGuardConfig` — shared JSON config (`nemoURL`, `timeoutSeconds`)
- `nemoResponse` / `nemoRailStatusEntry` — shared NeMo response types
- `callNemoGuard()` — shared HTTP call logic with `io.LimitReader` (CWE-400 mitigation), fail-closed status check, structured error reporting with per-rail detail in logs

Both plugins embed `nemoGuardBase` via Go composition, keeping each plugin file focused on its role-specific logic (message extraction and payload construction).

**Modified: `pkg/plugins/nemo/request_guard.go`**

Refactored to embed `nemoGuardBase` and delegate HTTP logic to `callNemoGuard()`. No behavioral changes.

**New file: `pkg/plugins/nemo/response_guard_test.go`**

Comprehensive tests covering:
- Construction and factory (valid config, missing fields, invalid JSON)
- `ProcessResponse` — allow, block (multiple scenarios), error (HTTP 500, invalid JSON), no-op (nil response, empty body, no choices, empty content)
- Payload verification (correct role, content, model forwarding)
- Trailing slash in `nemoURL`
- Fail-closed behavior (`"status": "allowed"` is rejected — only `"success"` passes)

**Modified: `pkg/plugins/nemo/request_guard_test.go`**

[...]

PR #130, opened 2026-04-15
refactor: update nemo-request-guard to use a shared nemo guard base

liavweiss:
## Summary

Extract shared HTTP/config logic from `nemo-request-guard` into a new `nemo_guard.go` file (`nemoGuardBase` struct), preparing for the upcoming `nemo-response-guard` plugin which will reuse the same base.

- **New file `nemo_guard.go`**: shared types (`nemoGuardConfig`, `nemoResponse`, `nemoRailStatusEntry`), constants, `nemoGuardBase` struct with `callNemoGuard` method
- **Refactored `request_guard.go`**: embeds `nemoGuardBase` instead of duplicating HTTP client, config types, and NeMo communication logic
- **Sanitized logging**: `callNemoGuard` logs only the host portion of the NeMo URL, not the full URL (avoids leaking credentials if URL contains userinfo or query tokens) - something that coderabbit catches in the response guard old PR (https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/113#discussion_r3059437830)

### Minor behavioral change

The 403 error message changed from `"request blocked by NeMo guardrails"` to `"nemo-request-guard: blocked by NeMo guardrails"` — the plugin type prefix is added by the shared `callNemoGuard` method to identify which guard plugin blocked the request (useful when both request and response guards are active).

## Test plan

- [x] All 26 unit tests pass (`go test ./pkg/plugins/nemo/`)
- [x] 2 new URL validation tests added
- [x] `go vet` clean
- [x] E2E tested on Kind cluster: Istio Gateway → Envoy ext-proc → BBR (refactored plugin) → NeMo guardrails
  - Safe requests → 200 (passed through to backend)
  - Bad request → 403 (blocked by NeMo)

## Summary by CodeRabbit

* **Refactor**
  * Internal reorganization of the NeMo request-guard plugin to introduce a shared guardrail base, consolidate NeMo request/response handling, centralize plugin naming, and simplify error routing.
* **Tests**
  * Updated unit test to reflect the new naming/initialization approach for the plugin instance.

### DIFF
--- a/pkg/plugins/nemo/nemo_guard_base.go
+++ b/pkg/plugins/nemo/nemo_guard_base.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2026 The opendatahub.io Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nemo
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	errcommon "sigs.k8s.io/gateway-api-inference-extension/pkg/common/error"
+	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/observability/logging"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
+)
+
+const (
+	nemoAllowedStatus    = "success"
+	defaultTimeoutSec    = 360
+	maxNemoResponseBytes = 1 << 20 // 1 MiB
+)
+
+// nemoGuardConfig is the shared JSON configuration for nemo guard plugins.
+type nemoGuardConfig struct {
+	NemoURL        string `json:"nemoURL"`
+	TimeoutSeconds int    `json:"timeoutSeconds"`
+}
+
+// nemoResponse is NeMo's JSON response from /v1/guardrail/checks.
+type nemoResponse struct {
+	Status      string                         `json:"status"`
+	RailsStatus map[string]nemoRailStatusEntry `json:"rails_status"`
+}
+
+type nemoRailStatusEntry struct {
+	Status string `json:"status"`
+}
+
+// nemoGuardBase holds the shared fields and HTTP logic for nemo guard plugins.
+type nemoGuardBase struct {
+	typedName  plugin.TypedName
+	nemoURL    string
+	httpClient *http.Client
+}
+
+func newNemoGuardBase(pluginType, nemoURL string, timeoutSeconds int) (*nemoGuardBase, error) {
+	if nemoURL == "" {
+		return nil, fmt.Errorf("nemoURL is required for plugin '%s'", pluginType)
+	}
+	timeout := time.Duration(timeoutSeconds) * time.Second
+	if timeout <= 0 {
+		timeout = defaultTimeoutSec * time.Second
+	}
+
+	return &nemoGuardBase{
+		typedName: plugin.TypedName{Type: pluginType, Name: pluginType},
+		nemoURL:   nemoURL,
+		httpClient: &http.Client{
+			Timeout: timeout,
+		},
+	}, nil
+}
+
+func (b *nemoGuardBase) TypedName() plugin.TypedName {
+	return b.typedName
+}
+
+// WithName sets the name of the plugin instance.
+func (b *nemoGuardBase) WithName(name string) *nemoGuardBase {
+	b.typedName.Name = name
+	return b
+}
+
+// callNemoGuard POSTs the payload to the NeMo guardrail checks endpoint, parses the
+// response, and returns an error with the corresponding error code if the content is
+// blocked or NeMo is unreachable. The caller is responsible for constructing the
+// client-facing errcommon.Error from the returned values.
+func (b *nemoGuardBase) callNemoGuard(ctx context.Context, payload []byte) (string, error) {
+	logger := log.FromContext(ctx)
+	logger.V(logutil.VERBOSE).Info("Calling NeMo guardrails", "pluginType", b.typedName.Type, "pluginName", b.typedName.Name, "url", b.nemoURL)
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, b.nemoURL, bytes.NewReader(payload))
+	if err != nil {
+		return errcommon.Internal, fmt.Errorf("create request: %w", err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	resp, err := b.httpClient.Do(httpReq)
+	if err != nil {
+		return errcommon.ServiceUnavailable, fmt.Errorf("call failed: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		return errcommon.ServiceUnavailable, fmt.Errorf("unexpected status %d", resp.StatusCode)
+	}
+
+	limited := io.LimitReader(resp.Body, maxNemoResponseBytes)
+	body, err := io.ReadAll(limited)
+	if err != nil {
+		return errcommon.ServiceUnavailable, fmt.Errorf("read response: %w", err)
+	}
+
+	var nemoResp nemoResponse
+	if err := json.Unmarshal(body, &nemoResp); err != nil {
+		return errcommon.ServiceUnavailable, fmt.Errorf("decode response: %w", err)
+	}
+
+	if strings.EqualFold(strings.TrimSpace(nemoResp.Status), nemoAllowedStatus) {
+		logger.V(logutil.VERBOSE).Info("Allowed by NeMo guardrails", "pluginType", b.typedName.Type, "pluginName", b.typedName.Name)
+		return "", nil
+	}
+
+	railsParts := make([]string, 0, len(nemoResp.RailsStatus))
+	for key, value := range nemoResp.RailsStatus {
+		railsParts = append(railsParts, fmt.Sprintf("%s: %s", key, value.Status))
+	}
+	railsStatus := fmt.Sprintf("[ %s ]", strings.Join(railsParts, " "))
+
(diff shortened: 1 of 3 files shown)


pkg/plugins/nemo/request_guard.go
nirrozenbaum: how about moving this function also to base?

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (b *nemoGuardBase) TypedName() plugin.TypedName {
+	return b.typedName
+}
+
+// callNemoGuard POSTs the payload to the NeMo guardrail checks endpoint, parses the
+// response, and returns an error if the content is blocked or NeMo is unreachable.
+func (b *nemoGuardBase) callNemoGuard(ctx context.Context, payload []byte) error {
+	logger := log.FromContext(ctx)
+	pluginType := b.typedName.Type
```
nirrozenbaum: we can remove this var `pluginType`. anyway that's not sufficient for logging. 
for log we need both type + name (nemo request guard type - is it mcp guard? inference guard? that will differ by different names) 

```suggestion
```

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	sanitizedHost := b.nemoURL
+	if u, err := url.Parse(b.nemoURL); err == nil {
+		sanitizedHost = u.Host
+	}
```
nirrozenbaum: I've read the PR intro about host sanitizing. this seems to me like overkill. this is the base nemo url. no users info or creds in the url. I think this can be removed.
 
```suggestion
```

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	logger.V(logutil.VERBOSE).Info("calling NeMo guardrails", "plugin", pluginType, "host", sanitizedHost)
```
nirrozenbaum: here we need also the plugin name.
```suggestion
	logger.V(logutil.VERBOSE).Info("Calling NeMo guardrails", "pluginType", b.TypedName.Type, "pluginName", b.TypedName.Name,  "url", b.nemoURL)
```

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, b.nemoURL, bytes.NewReader(payload))
+	if err != nil {
+		return errcommon.Error{Code: errcommon.Internal, Msg: fmt.Sprintf("%s: create request: %v", pluginType, err)}
```
nirrozenbaum: was writing initial comment, but then thinking about this again. 
I think the client should get back the plugins internal type/name. 
so we should remove plugin here.
maybe we should return back an error and the concrete plugin (e.g., request-guard or response guard) should create the error message that goes back to the client. without plugin type/name.
we can "help" the caller plugin by returning both (error, errcommon.Code)
```suggestion
		return err, errcommon.Internal
```

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	resp, err := b.httpClient.Do(httpReq)
+	if err != nil {
+		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: fmt.Sprintf("%s: call failed: %v", pluginType, err)}
```
nirrozenbaum: ditto

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: fmt.Sprintf("%s: unexpected status %d", pluginType, resp.StatusCode)}
```
nirrozenbaum: ditto

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	limited := io.LimitReader(resp.Body, maxNemoResponseBytes)
+	body, err := io.ReadAll(limited)
+	if err != nil {
+		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: fmt.Sprintf("%s: read response: %v", pluginType, err)}
```
nirrozenbaum: ditto

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	var nemoResp nemoResponse
+	if err := json.Unmarshal(body, &nemoResp); err != nil {
+		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: fmt.Sprintf("%s: decode response: %v", pluginType, err)}
```
nirrozenbaum: ditto

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if strings.EqualFold(strings.TrimSpace(nemoResp.Status), nemoAllowedStatus) {
+		logger.V(logutil.VERBOSE).Info("allowed by NeMo guardrails", "plugin", pluginType)
```
nirrozenbaum: plugin type is not enough to identify a plugin. 
we need plugin name as well.

```suggestion
		logger.V(logutil.VERBOSE).Info("allowed by NeMo guardrails", "pluginType", b.TypedName.Type, "pluginName", b.TypedName.Name)
```

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil
+	}
+
+	railsParts := make([]string, 0, len(nemoResp.RailsStatus))
+	for key, value := range nemoResp.RailsStatus {
+		railsParts = append(railsParts, fmt.Sprintf("%s: %s", key, value.Status))
+	}
+	railsStatus := fmt.Sprintf("[ %s ]", strings.Join(railsParts, " "))
+
+	logger.Info("blocked by NeMo guardrails", "plugin", pluginType, "railsStatus", railsStatus)
```
nirrozenbaum: ditto about plugin name

pkg/plugins/nemo/nemo_guard.go
```diff
@@ -0,0 +1,138 @@
...
+	return errcommon.Error{Code: errcommon.Forbidden, Msg: fmt.Sprintf("%s: blocked by NeMo guardrails", pluginType)}
```
nirrozenbaum: return error + status code

pkg/plugins/nemo/nemo_guard_base.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+const (
+	nemoAllowedStatus    = "success"
+	defaultTimeoutSec    = 360
+	maxNemoResponseBytes = 1 << 20 // 1 MiB
+)
+
+// nemoGuardConfig is the shared JSON configuration for nemo guard plugins.
+type nemoGuardConfig struct {
```
nirrozenbaum, line 42: I assume at later PRs we might want to add another flag to the config, something like (not sure about the name):
```yaml
APIFormat: "openai-chat-completions" | "mcp-jsonrpc" (with some default?)
```
and then parse the request accordingly.
so we would have a configurable nemo parser with two types built-in (for chat completions and mcp-jsonrpc). this is extensible in case we want to support in the future more formats (e.g., like responses API)
cc: @christinaexyou

not required for this PR.

pkg/plugins/nemo/nemo_guard_base.go
```diff
@@ -0,0 +1,138 @@
+/*
```
nirrozenbaum, line 1: maybe we should name the file `nemo_guard_base.go`. 
otherwise that's a bit confusing to understand what's the difference between `nemo_guard` and `nemo_request_guard`

pkg/plugins/nemo/request_guard.go
```diff
@@ -155,50 +107,7 @@ func (p *NemoRequestGuardPlugin) ProcessRequest(ctx context.Context, _ *framewor
...
-	// handle block message
-	railsParts := make([]string, 0, len(response.RailsStatus))
-	for key, value := range response.RailsStatus {
-		railsParts = append(railsParts, fmt.Sprintf("%s: %s", key, value.Status))
-	}
-	railsStatus := fmt.Sprintf("[ %s ]", strings.Join(railsParts, " "))
-
-	logger.Info("request blocked by NeMo guardrails", "railsStatus", railsStatus)
-	return errcommon.Error{Code: errcommon.Forbidden, Msg: "request blocked by NeMo guardrails"}
+	return p.callNemoGuard(ctx, payload)
```
nirrozenbaum: if we get back err + status code, we need to change here accordingly.

pkg/plugins/nemo/request_guard_test.go
```diff
@@ -88,7 +88,7 @@ func TestNemoRequestGuardTypedName(t *testing.T) {
 // --- ProcessRequest: allow / block / error ---
 
 func TestNemoRequestGuardProcessRequest(t *testing.T) {
-	const forbiddenMsg = "request blocked by NeMo guardrails"
+	const forbiddenMsg = "nemo-request-guard: blocked by NeMo guardrails"
```
nirrozenbaum: let's try to avoid internal plugin names in client returned messages. this tends to break quickly.

pkg/plugins/nemo/nemo_guard_base.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (b *nemoGuardBase) TypedName() plugin.TypedName {
+	return b.typedName
+}
+
+// WithName sets the name of the plugin instance.
+func (b *nemoGuardBase) WithName(name string) {
+	b.typedName.Name = name
+}
```
nirrozenbaum: align with the `WithName` implementation of all other plugins
```suggestion
func (b *nemoGuardBase) WithName(name string) *nemoGuardBase {
	b.typedName.Name = name
	return b
}
```

liavweiss: sure thanks

pkg/plugins/nemo/request_guard.go
```diff
@@ -89,38 +56,18 @@ func NemoRequestGuardFactory(name string, rawParameters json.RawMessage, _ frame
 		return nil, fmt.Errorf("failed to create '%s' plugin - %w", NemoRequestGuardPluginType, err)
 	}
 
-	return p.WithName(name), nil
+	p.WithName(name)
+	return p, nil
```
nirrozenbaum:
```suggestion
	return p.WithName(name), nil
```

nirrozenbaum (approved):
/lgtm
/approve

Thanks!

PR #138, opened 2026-04-18
fix: use sslip.io FQDN for llm-katan simulator endpoint

yossiovadia:
## Summary
- Replace bare IP (`3.13.21.181`) with sslip.io FQDN (`3-13-21-181.sslip.io`) for the llm-katan test simulator endpoint
- `sslip.io` is a free DNS service that resolves `X-X-X-X.sslip.io` → `X.X.X.X`, so connectivity is unchanged
- The simulator now has a Let's Encrypt TLS certificate for the sslip.io domain, enabling proper cert verification in Istio without `insecureSkipVerify`

## Why
- Bare IPs are rejected by community Istio's ServiceEntry validation (`resolution: DNS` requires a hostname)
- Self-signed certs cause `CERTIFICATE_VERIFY_FAILED` with `tls.mode: SIMPLE` in Envoy
- Let's Encrypt blocks `*.amazonaws.com` domains by policy, so EC2 public DNS names can't get certs
- `sslip.io` solves both: valid FQDN for Istio + trusted TLS cert from Let's Encrypt

## Test plan
- [ ] CI E2E tests pass with the new FQDN
- [ ] `curl https://3-13-21-181.sslip.io/health` returns 200 with valid cert
- [ ] `curl http://3-13-21-181.sslip.io:8000/health` returns 200 (HTTP still works)

## Summary by CodeRabbit

* **Chores**
  * Updated test simulator endpoint configuration from IP address to DNS hostname across CI/CD workflows and test infrastructure scripts.

### DIFF
--- a/.github/workflows/ci-e2e.yaml
+++ b/.github/workflows/ci-e2e.yaml
@@ -9,7 +9,7 @@ on:
       - main
 
 env:
-  E2E_SIMULATOR_ENDPOINT: "3.13.21.181"
+  E2E_SIMULATOR_ENDPOINT: "3-13-21-181.sslip.io"
 
 jobs:
   check-changes:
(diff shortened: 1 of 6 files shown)


yossiovadia:
Closing — the BBR E2E tests use `resolution: STATIC` with a fake hostname and bare IP, which is a different routing pattern from the MaaS ExternalModel controller. The FQDN change breaks the test infrastructure because Caddy rejects TLS connections where the SNI doesn't match its configured domain (unlike the old self-signed llm-katan which accepted any SNI). The Let's Encrypt cert on the simulator is still valuable for production use via the MaaS/ExternalModel path.

PR #140, opened 2026-04-18
fix: use sslip.io FQDN for llm-katan simulator endpoint

yossiovadia:
## Summary
- Replace bare IP (`3.13.21.181`) with sslip.io FQDN (`3-13-21-181.sslip.io`)
- The llm-katan simulator now uses Caddy + Let's Encrypt TLS on port 443, replacing the old self-signed cert
- Caddy requires proper SNI matching — bare IP connections are rejected

## Changes
- `resolution: STATIC` → `resolution: DNS` (FQDN instead of bare IP)
- Fake hostname `e2e-simulator.external` → real FQDN from `simulatorEP`
- Added `sni` field to DestinationRule (Istio leaves SNI empty when `insecureSkipVerify: true` unless explicitly set)
- Kept `insecureSkipVerify: true` (matches ExternalModel controller behavior)

## Test plan
- [x] Verified locally: 5/5 connectivity tests pass (should return 200)
- [ ] 5 response body format tests fail — this is a pre-existing ext-proc response body handling issue with Caddy's chunked transfer encoding, unrelated to this change. Will investigate separately.

## Why this is urgent
The simulator's TLS was changed from self-signed to Let's Encrypt (Caddy on port 443). Without this PR, the E2E CI is broken on `main` — bare IP connections to Caddy get `TLSV1_ALERT_INTERNAL_ERROR`.

## Summary by CodeRabbit

* **Chores**
  * Updated test and CI/CD infrastructure configuration to use DNS-based simulator endpoints with corresponding service mesh adjustments for improved reliability.

### DIFF
--- a/.github/workflows/ci-e2e.yaml
+++ b/.github/workflows/ci-e2e.yaml
@@ -9,7 +9,7 @@ on:
       - main
 
 env:
-  E2E_SIMULATOR_ENDPOINT: "3.13.21.181"
+  E2E_SIMULATOR_ENDPOINT: "3-13-21-181.sslip.io"
 
 jobs:
   check-changes:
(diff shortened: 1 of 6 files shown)


yossiovadia:
Closing for now — reverted 3.13.21.181 to self-signed so CI passes on main. Investigating the ext-proc response body issue against a new Caddy instance (16.59.49.55 / 16-59-49-55.sslip.io) separately.

PR #142, opened 2026-04-19
fix(e2e): drop insecureSkipVerify and use Let's Encrypt simulator endpoint

### LINKED ISSUE #141: E2E: drop insecureSkipVerify now that simulator uses Let's Encrypt certs
## Context

The llm-katan simulator instances now use publicly trusted Let's Encrypt
certificates via sslip.io DNS (e.g., `3-13-21-181.sslip.io`), replacing
the previous self-signed certificates on raw IPs.

## Problem

The E2E test DestinationRule sets `insecureSkipVerify: true` as a workaround
for the simulator's self-signed certificates. This is no longer needed and
doesn't match production behavior where real providers (OpenAI, Anthropic,
Bedrock) use publicly trusted certs.

## Fix

- Remove `insecureSkipVerify: true` from the E2E DestinationRule
- Switch ServiceEntry from `STATIC` (IP-based) to `DNS` resolution
- Update default simulator endpoint from `3.13.21.181` to `3-13-21-181.sslip.io`
- Update all scripts and CI workflow references

Verified on RHOAI cluster: all simulator models work with `tls.mode: SIMPLE`
and no `insecureSkipVerify`, matching real provider behavior.

Related: opendatahub-io/ai-gateway-payload-processing#138

noyitz:
## Summary

- Remove `insecureSkipVerify: true` from E2E DestinationRule
- Switch simulator from raw IP (`3.13.21.181`) to DNS (`3-13-21-181.sslip.io`) with Let's Encrypt cert
- Switch ServiceEntry from `STATIC` to `DNS` resolution
- Update ExternalName Services to point to DNS directly
- Update all default endpoints in scripts and CI workflow

Fixes #141

## Why

The llm-katan simulator now uses publicly trusted Let's Encrypt certificates.
The `insecureSkipVerify` workaround is no longer needed and doesn't match
production behavior (real providers use trusted certs).

## Test Plan

- [x] Verified on RHOAI cluster: all 3 simulator models (sim-openai, sim-anthropic, sim-bedrock) work with `tls.mode: SIMPLE`, no `insecureSkipVerify`
- [x] Streaming works through Let's Encrypt endpoint
- [x] `go vet ./test/e2e/` passes
- [x] No remaining references to old IP or `insecureSkipVerify` in test code

## Summary by CodeRabbit

* **Chores**
  * Updated CI and setup configuration to use a DNS-based simulator endpoint hostname instead of a fixed IP for improved connectivity and clarity.

* **Tests**
  * End-to-end test manifests and scripts now parameterize the simulator endpoint and use DNS resolution with updated TLS-related settings to better match real-world simulator connectivity.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -21,7 +21,7 @@ import (
 
 const (
 	defaultNs                = "bbr-e2e"
-	defaultSimulatorEndpoint = "3.13.21.181"
+	defaultSimulatorEndpoint = "3-13-21-181.sslip.io"
 	defaultGatewayName       = "e2e-gateway"
 	defaultGatewayNamespace  = "default"
 )
@@ -74,15 +74,13 @@ metadata:
   namespace: %s
 spec:
   hosts:
-  - e2e-simulator.external
+  - %s
   location: MESH_EXTERNAL
   ports:
   - number: 443
     name: https
     protocol: HTTPS
-  resolution: STATIC
-  endpoints:
-  - address: %s
+  resolution: DNS
 ---
 apiVersion: networking.istio.io/v1
 kind: DestinationRule
@@ -90,12 +88,11 @@ metadata:
   name: e2e-simulator
   namespace: %s
 spec:
-  host: e2e-simulator.external
+  host: %s
   trafficPolicy:
     tls:
       mode: SIMPLE
-      insecureSkipVerify: true
-`, nsName, simulatorEP, nsName))
+`, nsName, simulatorEP, nsName, simulatorEP))
 
 	ginkgo.By("Creating curl client pod")
 	kubectlApplyLiteral(fmt.Sprintf(`
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -92,7 +92,9 @@ spec:
   trafficPolicy:
     tls:
       mode: SIMPLE
-`, nsName, simulatorEP, nsName, simulatorEP))
+      sni: %s
+      caCertificates: /etc/ssl/certs/ca-certificates.crt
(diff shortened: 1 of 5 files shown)


nirrozenbaum:
@noyitz e2e seems to fail..

noyitz:
> @noyitz e2e seems to fail..

@nirrozenbaum  fixed

test/e2e/e2e_suite_test.go
```diff
@@ -74,28 +74,27 @@ metadata:
...
 spec:
-  host: e2e-simulator.external
+  host: %s
   trafficPolicy:
     tls:
       mode: SIMPLE
-      insecureSkipVerify: true
-`, nsName, simulatorEP, nsName))
+      sni: %s
+      caCertificates: /etc/ssl/certs/ca-certificates.crt
```
nirrozenbaum, line 96: this part is not implemented in the real reconciler in maas. it won't be included in real integration with the ExternalModel CR.

nirrozenbaum:
@noyitz this PR will work for simulator based e2e tests. 
but the PR is diverging from what the ExternalModel reconciler is doing (the reconciler is not setting sni nor caCertificates). 

if we're not testing what's the reconciler is doing, that's not good. 
(another good opportunity to say we should move the reconciler to this repo).

can we fix this in the real reconciler? do we know if `caCertificates: /etc/ssl/certs/ca-certificates.crt` is working with all real providers? 

until this change is updated in the reconciler and tested to work with real providers, we cannot merge this change IMO (I personally would defer this change to post 3.4)

noyitz:
since we can't match the reconciler's DestinationRule config in Kind CI (due to Istio's missing CA), we keep insecureSkipVerify for the E2E CI tests.

PR #146, opened 2026-04-19
fix: return 400 for empty messages in openai, anthropic, bedrock translators

### LINKED ISSUE #144: Anthropic translator returns 500 on empty messages array instead of 400
## Problem

Sending `"messages": []` to an Anthropic-provider ExternalModel returns HTTP 500 instead of 400.

The Anthropic translator validates that at least one non-system message exists (line 63 in `anthropic.go`) and returns an error. The BBR framework converts this plugin error into a 500 Internal Server Error. The expected behavior is 400 Bad Request.

## Reproduction

```bash
curl -sk "https://${GATEWAY_HOST}/llm/<anthropic-model>/v1/chat/completions" \
  -H "Authorization: Bearer ${API_KEY}" \
  -H "Content-Type: application/json" \
  -d '{"model":"<target-model>","messages":[]}'
```

Returns: HTTP 500 (empty body)
Expected: HTTP 400 with error message "at least one non-system message is required"

## Root Cause

The translator correctly returns an error, but the BBR ext_proc framework maps all plugin errors to 500. Either:
1. The framework should support returning 4xx for validation errors
2. Or the translator should handle this case before returning an error (e.g., pass through to the provider which will return its own 400)

## Affects

Both real Anthropic API and llm-katan simulator — confirmed on RHOAI cluster.

noyitz:
## Summary

All three translators now validate the messages array and return HTTP 400 instead of 500
for empty/missing messages.

Fixes #144

## Problem

Sending `"messages": []` returned HTTP 500 because translator errors were plain `fmt.Errorf`,
which the BBR framework maps to `Unknown` → gRPC error → 500. The framework supports
`errcommon.Error{Code: BadRequest}` which maps to HTTP 400 ImmediateResponse.

## Changes

| Translator | Before | After |
|-----------|--------|-------|
| openai | No messages validation — passed through to provider (503) | Validates messages, returns 400 |
| anthropic | Validated but returned `fmt.Errorf` → 500 | Returns `errcommon.Error{Code: BadRequest}` → 400 |
| bedrock-openai | No messages validation — passed through to provider (503) | Validates messages, returns 400 |

## Test Plan

- [x] `TestTranslateRequest_EmptyMessages` — openai (new test file)
- [x] `TestTranslateRequest_MissingMessages` — openai (new test file)
- [x] `TestTranslateRequest_EmptyMessages` — anthropic (existing, still passes)
- [x] `TestTranslateRequest_EmptyMessages` — bedrock (new test)
- [x] `TestTranslateRequest_MissingMessages` — bedrock (new test)
- [x] Full unit test suite passes (`go test ./pkg/...`)

### DIFF
--- a/pkg/plugins/api-translation/plugin.go
+++ b/pkg/plugins/api-translation/plugin.go
@@ -19,9 +19,11 @@ package api_translation
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/bbr/framework"
+	errcommon "sigs.k8s.io/gateway-api-inference-extension/pkg/common/error"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
 
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/api-translation/translator"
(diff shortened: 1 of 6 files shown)


nirrozenbaum (approved):
/lgtm
/approve

PR #133, opened 2026-04-15
feat: add Vertex AI OpenAI-compatible (vertex-openai) api-translation provider

asaadbalum:
## Summary

Adds a new `vertex-openai` provider that targets Vertex AI Platform's native OpenAI-compatible `chat/completions` endpoint ([docs](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.endpoints.chat/completions)). Unlike the existing `vertex` translator (which converts OpenAI format to Gemini `generateContent` format), this translator passes the request body through unchanged since the endpoint accepts OpenAI format natively.

Closes #112

## Changes

- **New translator**: `vertex_openai.go` — passthrough translator that constructs the full Vertex AI path `/v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions` from GCP project, location, and endpoint parameters
- **Response stripping**: Strips `usage.extra_properties` (Vertex AI-specific field not in OpenAI spec), following Azure's `ResponseFieldStripper` pattern
- **Validation**: `model` and `messages` fields validated using `errcommon.Error`
- **Provider registration**: `vertex-openai` constant in `provider.go`
- **Plugin wiring**: Registered in api-translation and apikey-injection plugins (Bearer token auth, OAuth2)
- **E2E**: `vertex-openai` entry added to the E2E provider list

## Test Plan

- [x] Unit tests: 8 test cases (5 request: passthrough, missing/empty model, missing/empty messages; 3 response: no-op, extra_properties stripping, error passthrough)
- [x] Integration: `golangci-lint run ./...` — 0 issues
- [x] Real endpoint: tested against `aiplatform.googleapis.com/v1/projects/.../endpoints/openapi/chat/completions` — HTTP 200, verified response format and identified `usage.extra_properties` for stripping
- [x] All 14 packages pass (`go test`)

## Note: Vertex AI Platform vs Generative Language API

This PR targets Vertex AI Platform (`aiplatform.googleapis.com`) per issue #112. The Generative Language API (`generativelanguage.googleapis.com`) uses a different path (`/v1beta/openai/chat/completions`) and is tracked separately in issue #143.

## Follow-up

Config wiring to inject project/location/endpoint from the ExternalModel CRD into the translator constructor (currently initialized with empty defaults in `plugin.go`).

## Summary by CodeRabbit

* **New Features**
  * Added support for the Vertex OpenAI provider: request/response translation, routing to Vertex endpoints, and automatic API key / auth header handling.
[...]

PR #151, opened 2026-04-21
fix(e2e): always build BBR image from source for E2E tests

noyitz:
## Problem

The E2E CI setup uses the pre-built `quay.io/opendatahub/odh-ai-gateway-payload-processing:odh-stable`
image, which doesn't include changes from the PR being tested. This means new providers, translators,
or bug fixes in a PR are never actually tested in E2E — the tests run against the old released binary.

This was the root cause of the `vertex-openai` timeout in PR #133: the BBR pod had the config args
for `vertex-openai`, but the binary didn't have the translator code because it was using the old image.

## Fix

Always build the Docker image from the PR's source code and load it into Kind:

```bash
docker build -t "$image_tag" --platform "$platform" "$PROJECT_ROOT"
kind load docker-image "$image_tag" --name "$KIND_CLUSTER_NAME"
```

This replaces the previous approach which only built from source on arm64 Macs and used the
pre-built quay.io image on x86 CI (GitHub Actions).

## Impact

- CI build time increases by ~1-2 minutes (Docker build step)
- E2E tests now actually validate the PR's code changes
- Works on both x86 (CI) and arm64 (Mac) environments

## Summary by CodeRabbit

* **Chores**
  * Updated internal test environment setup scripts to improve multi-architecture build support.

### DIFF
--- a/test/e2e/scripts/setup-kind.sh
+++ b/test/e2e/scripts/setup-kind.sh
@@ -117,15 +117,17 @@ EOF
 deploy_bbr() {
     echo "Deploying BBR via Helm chart..."
 
-    # Build and load image for arm64 if on Mac (quay.io image is x86 only)
+    local image_tag="quay.io/opendatahub/odh-ai-gateway-payload-processing:odh-stable"
+
+    # Always build from source to ensure E2E tests run against the current code.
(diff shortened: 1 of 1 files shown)


yossiovadia (approved):
LGTM

PR #152, opened 2026-04-21
api translation plugin init

nirrozenbaum:
## Merge criteria:

## Summary by CodeRabbit

* **Enhancements**
  * VertexOpenAI configuration is now optional, allowing setups to omit that key when unused.
  * Initialization now emits more detailed logs showing which providers are configured for improved visibility.

* **Refactor**
  * Plugin initialization and construction now use the provided context/handle to improve lifecycle and resource management.

### DIFF
--- a/pkg/plugins/api-translation/plugin.go
+++ b/pkg/plugins/api-translation/plugin.go
@@ -104,7 +104,7 @@ func NewAPITranslationPlugin(ctx context.Context, config apiTranslationConfig) (
 		keys = append(keys, key)
 	}
 
-	log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized with providers [ %s ]", strings.Join(keys, ","))
+	log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized", "providers", strings.Join(keys, ","))
 
 	return &APITranslationPlugin{
 		typedName: plugin.TypedName{
--- a/pkg/plugins/api-translation/plugin.go
+++ b/pkg/plugins/api-translation/plugin.go
@@ -21,9 +21,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/bbr/framework"
 	errcommon "sigs.k8s.io/gateway-api-inference-extension/pkg/common/error"
+	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/observability/logging"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
 
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/api-translation/translator"
@@ -46,7 +49,7 @@ var _ framework.ResponseProcessor = &APITranslationPlugin{}
 
 // apiTranslationConfig holds configuration for provider-specific translators.
 type apiTranslationConfig struct {
-	VertexOpenAI *vertexOpenAIConfig `json:"vertexOpenAI"`
+	VertexOpenAI *vertexOpenAIConfig `json:"vertexOpenAI,omitempty"`
 }
 
 type vertexOpenAIConfig struct {
@@ -56,7 +59,7 @@ type vertexOpenAIConfig struct {
 }
 
 // APITranslationFactory defines the factory function for APITranslationPlugin.
-func APITranslationFactory(name string, rawConfig json.RawMessage, _ framework.Handle) (framework.BBRPlugin, error) {
+func APITranslationFactory(name string, rawConfig json.RawMessage, handle framework.Handle) (framework.BBRPlugin, error) {
 	var config apiTranslationConfig
 	if len(rawConfig) > 0 {
 		if err := json.Unmarshal(rawConfig, &config); err != nil {
@@ -64,7 +67,7 @@ func APITranslationFactory(name string, rawConfig json.RawMessage, _ framework.H
 		}
 	}
 
-	p, err := NewAPITranslationPlugin(config)
+	p, err := NewAPITranslationPlugin(handle.Context(), config)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


yossiovadia:
## OpinAI PR Review

**PR:** #152
**Title:** PR #152
**Author:** @nirrozenbaum
**Verdict:** CHANGES_REQUESTED
**Risk:** MEDIUM

Now I have full context. I can see that `apikey-injection/plugin.go` at line 145 uses the correct `logr` structured logging pattern: `.Info("message string", "key", value)` — key-value pairs, not `fmt.Sprintf` style.

The PR's logging at line 107 uses:
```go
log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized with providers [ %s ]", strings.Join(keys, ","))
```

This is **incorrect** for `logr.Logger.Info()`. The `logr.Info` signature is `Info(msg string, keysAndValues ...interface{})`. The `%s` won't be interpolated — it will appear literally in the log message, and `strings.Join(keys, ",")` will be treated as a key without a corresponding value (odd number of key-value args), which will produce malformed log output.

Now I have enough to deliver my review.

---

--- OPINAI PR REVIEW ---

## Summary

This PR makes three changes to the `api-translation` plugin:
1. Makes the `VertexOpenAI` config field optional in JSON serialization (adds `omitempty` tag)
2. Passes `context.Context` through to `NewAPITranslationPlugin` via `handle.Context()` for logging purposes
3. Adds verbose logging at plugin initialization to display all configured providers

## Risk Assessment

**MEDIUM** — Contains a confirmed bug in the logging call that will produce malformed log output, and an existing test that will now panic due to a nil `handle` dereference.

## Previously Identified

- @coderabbitai already flagged the **nil `handle` dereference** risk in `APITranslationFactory` at line 62-70, where `handle.Context()` is called without a nil check. They noted the existing test at `plugin_test.go:532` passes `nil` as the handle: `APITranslationFactory("test-instance", nil, nil)`.
- @coderabbitai already flagged the **incorrect `logr.Info()` usage** with `%s` format placeholder at `plugin.go:107`.

## New Findings

### 1. 🐛 [BUG] Incorrect `logr.Logger.Info()` call — detailed analysis (plugin.go:107)

While @coderabbitai flagged this, I want to provide a precise explanation of the severity:

```go
log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized with providers [ %s ]", strings.Join(keys, ","))
```

The `logr.Logger.Info()` signature is `Info(msg string, keysAndValues ...interface{})`. This means:
- The message will literally contain `%s` — it is **not** `fmt.Sprintf`-style formatting
- `strings.Join(keys, ",")` will be treated as a **key** in the key-value pairs, with **no corresponding value** (odd number of args)
- Most logr implementations (like `zapr`) will log a malformed entry with an error about odd key-value pairs

**Correct form** (following the pattern at `apikey-injection/plugin.go:145`):
```go
log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized with providers", "providers", strings.Join(keys, ","))
```

### 2. 🐛 [BUG] Existing test `TestAPITranslationFactory_NilConfig` will panic (plugin_test.go:532)

From grep results, `plugin_test.go:532` contains:
```go
p, err := APITranslationFactory("test-instance", nil, nil)
```

Before this PR, the `handle` parameter was `_` (unused), so passing `nil` was safe. Now that `handle.Context()` is called on line 70, this test will panic with a nil pointer dereference. This is a **test regression** that should be caught by CI.

### 3. Minor: Non-deterministic log output (plugin.go:103-106)

The providers map is iterated with `range`, which in Go has non-deterministic order. This means the log message will list providers in a different order each time. While this is a minor cosmetic issue, for log consistency consider sorting the keys:

```go
keys := make([]string, 0, len(providers))
for key := range providers {
    keys = append(keys, key)
}
sort.Strings(keys)
```

### 4. Minor: `omitempty` on `VertexOpenAI` (plugin.go:52)

Adding `omitempty` to the JSON tag is fine for marshaling, but note it has **no effect on unmarshaling** (which is the only direction used in this code — config is parsed from `rawConfig`). This change is harmless but cosmetic only in this context.

## Suggestions

1. **Fix the `.Info()` call** to use proper structured logging:
   ```go
   log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized", "providers", keys)
   ```
   (Passing the `[]string` slice directly is also valid for most logr backends.)

2. **Fix the factory test** to pass a mock/real handle instead of `nil`, or add a nil guard in `APITranslationFactory`:
   ```go
   ctx := context.Background()
   if handle != nil {
       ctx = handle.Context()
   }
   ```

3. Consider if `NewAPITranslationPlugin` truly needs a `context.Context` parameter for a single log line during initialization. This is a one-time startup log — passing `context.Background()` internally could avoid the API change and the nil-handle risk entirely.

## Test Coverage

- The `newTestPlugin()` helper was correctly updated to pass `context.Background()`.
- However, the existing `TestAPITranslationFactory_NilConfig` test at line 532 was **not** updated and will panic with the new code, indicating the test suite was not fully run against this change.

## Verdict

**CHANGES_REQUESTED** — The logging call is incorrect (will produce malformed log output) and the existing factory test will panic. Both issues need to be fixed before merge.

--- END PR REVIEW ---

===PR_VERDICT===
verdict: CHANGES_REQUESTED
risk: MEDIUM
===END_PR_VERDICT===

--- OPINAI SUGGESTED_QUESTIONS ---
["In plugin_test.go:532, `APITranslationFactory(\"test-instance\", nil, nil)` passes nil as the handle — did CI pass with this change, or was this test not run?", "The logr `.Info()` method uses structured key-value pairs, not printf formatting. Was the intent to log providers as a key-value pair like `\"providers\", keys` rather than using `%s` interpolation?", "Is there value in having `NewAPITranslationPlugin` accept a `context.Context` just for a single initialization log line? Could this be done internally with `context.Background()` to keep the API simpler?", "Should the provider names in the log output be sorted for deterministic, reproducible log lines across restarts?", "How does the framework guarantee that `handle` is non-nil when calling factory functions — is there a contract in the `framework.Register` path that ensures this, or should factories defensively handle nil?"]
--- END SUGGESTED_QUESTIONS ---

pkg/plugins/api-translation/plugin.go
```diff
@@ -101,6 +99,13 @@ func NewAPITranslationPlugin(config apiTranslationConfig) (*APITranslationPlugin
 		)
 	}
 
+	keys := make([]string, 0, len(providers))
+	for key := range providers {
+		keys = append(keys, key)
+	}
+
+	log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized with providers [ %s ]", strings.Join(keys, ","))
+
```
noyitz, line 108: The Info method in logr doesn't use %s formatting — the second arg is a key-value pair, not a printf arg

`log.FromContext(ctx).V(logutil.VERBOSE).Info("plugin initialized", "providers", strings.Join(keys, ","))`

nirrozenbaum:
@noyitz @yossiovadia thanks. remind me not to push PRs after not sleeping 2 days 😆 .
should be fixed now

PR #161, opened 2026-04-23
fix(e2e): install ExternalModel CRD in entrypoint if not present

noyitz:
The E2E test container entrypoint now checks for the ExternalModel CRD
and installs it if missing. This fixes the shift-left Jenkins pipeline
where the cluster may not have MaaS fully deployed.

The GitHub Actions CI doesn't hit this because `setup-kind.sh` installs
the CRD during cluster setup. Jenkins runs the container directly via
`podman run` without the setup script.

## Summary by CodeRabbit

* **Tests**
  * Enhanced E2E test setup to ensure required CustomResourceDefinition is available before test execution.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -37,7 +37,7 @@ var (
 
 func TestE2E(t *testing.T) {
 	gomega.RegisterFailHandler(ginkgo.Fail)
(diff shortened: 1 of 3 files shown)


nirrozenbaum (approved):
/lgtm
/approve

PR #162, opened 2026-04-23
fix: forward full conversation context to NeMo, filter unsupported system role

### LINKED ISSUE #160: security issue in nemo guard plugin
The NeMo Request Guard only extracts and validates the LAST user message from the conversation, completely ignoring: - System/developer messages (can inject arbitrary instructions) - Tool response messages (can contain embedded prompt injection) - Earlier user messages in conversation history
Attack Vectors: 
1. System message injection: Malicious instructions in "role": "system" bypass guardrails entirely 
2. Tool response poisoning: Prompt injection embedded in "role": "tool" content forwarded without validation

Impact: 
- Complete guardrail bypass via non-user message roles 
- Multi-turn conversation exploits through unvalidated history

References: OWASP LLM Top 10 – LLM01 (Prompt Injection), LLM08 (Excessive Agency), CWE-20

liavweiss:
## Summary

Fixes #160 - the request guard previously only sent the **last user message** to NeMo for
input-rail evaluation. This allowed toxic content in earlier user turns, assistant messages,
and tool messages to bypass guardrails entirely.

### Changes

- **`extractMessages`** now forwards all messages (user, assistant, tool) to NeMo instead of
  only the last user message.
- **System messages are filtered out** because NeMo's `/v1/guardrail/checks` endpoint
  (TrustyAI fork) rejects them with `status: "error"` - confirmed with TrustyAI team (Christina).
  See: [server/api.py#L1041](https://github.com/trustyai-explainability/NeMo-Guardrails/blob/6a3cfd1e51fd978075adc4355846dcea5927d60c/nemoguardrails/server/api.py#L1041)
- Updated `ProcessRequest` doc comment.
- Updated and added unit tests:
  - `"multi-turn conversation — all messages forwarded"`
  - `"system message filtered out — /v1/guardrail/checks rejects system role"`
  - `"system-only conversation — all filtered, returns nil"`
  - `"tool message included — not filtered out"`
  - `TestNemoRequestGuardForwardsAllNonSystemMessages` — integration test verifying system
    is excluded while user/assistant/tool are forwarded to NeMo.

### Testing

- All unit tests pass (`go test ./pkg/plugins/nemo/...`)
- E2E verified on Kind cluster:
  - Safe request with system + user → **200 OK** (system filtered, user evaluated)
  - Toxic earlier user message with safe follow-up → **403 Forbidden** (previously passed)
  - Toxic user-only message → **403 Forbidden**

## Summary by CodeRabbit

* **Bug Fixes**
  * Forward all non-system messages (user, assistant, tool) in original order instead of only the last user message.
  * Filter out system messages from forwarded requests; system-only conversations now yield no forwarded messages.

* **Documentation**
  * Updated docs to reflect the new message extraction and system-message filtering behavior.

* **Tests**
  * Added/updated tests covering multi-turn ordering, system-message filtering, and inclusion of tool-role messages.

### DIFF
--- a/pkg/plugins/nemo/request_guard.go
+++ b/pkg/plugins/nemo/request_guard.go
@@ -133,7 +133,7 @@ func (p *NemoRequestGuardPlugin) ProcessRequest(ctx context.Context, _ *framewor
 // extractMessages returns user-supplied text as a message slice suitable for NeMo's
 // OpenAI-compatible chat endpoint. It supports two payload formats:
 //
-//  1. OpenAI chat: top-level "messages" array → returns the last user message.
+//  1. OpenAI chat: top-level "messages" array → forwards all non-system messages.
 //  2. MCP JSON-RPC: {"jsonrpc":"2.0","params":{"arguments":{…}}} → concatenates
 //     all string argument values into a single user message.
 //
@@ -148,8 +148,9 @@ func extractMessages(body map[string]any) ([]map[string]string, error) {
 	return nil, nil // not an inference request (e.g. API key management, model listing)
 }
 
-// extractOpenAIMessages parses an OpenAI-style "messages" value and returns the
-// last user message, or all messages as a fallback when no user message exists.
+// extractOpenAIMessages parses an OpenAI-style "messages" value. System messages are
+// filtered. TrustyAI does not support this role. All other roles are forwarded so NeMo can evaluate
+// the full conversation context.
 func extractOpenAIMessages(raw any) ([]map[string]string, error) {
 	slice, ok := raw.([]any)
 	if !ok {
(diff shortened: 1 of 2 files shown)


nirrozenbaum:
@liavweiss can you please rebase?

pkg/plugins/nemo/request_guard.go
```diff
@@ -163,18 +164,13 @@ func extractOpenAIMessages(raw any) ([]map[string]string, error) {
 			continue
 		}
 		role, _ := msg["role"].(string)
+		if role == "system" {
```
nirrozenbaum, line 167: can an attacker use system role to inject malicious content?

liavweiss: That is a good point - you are right, an attacker could put malicious content in a system role message and bypass the input guardrails. I filtered system messages because the TrustyAI NeMo endpoint rejects them with a hard error. 

Maybe we can remap `system` to `user` before sending to NeMo, both have trade-offs (remapping could cause false positives on legitimate system prompts, e.g., "Never discuss violence" flagged as toxic user input, filtering leaves a security gap), but we would probably rather handle false positives than leave a potential security bypass.

What do you think?

nirrozenbaum: mmmm. false positives due to our plugin is not a good choice :). 
I'd say we should probably go with checking all content other than "system" messages and in parallel work with Trusty team to remove this limitation. that would be an incremental progress towards the right solution. 

just need to make sure we keep the original issue opened and fix it.
timeline wise, we have by next release to fix it, so we have time (but ideally the sooner the better).

nirrozenbaum: resolving the comment for now, as your code is aligned with filtering only system messages.

nirrozenbaum (approved):
/lgtm
/approve

nice work 🥳

PR #188, opened 2026-04-27
fix: pass response_format through Anthropic translator

### LINKED ISSUE #186: bug: Anthropic translator strips response_format, breaking JSON mode
## Description

The Anthropic api-translation plugin strips the `response_format` field when translating OpenAI-format requests to Anthropic's native format. This means JSON mode (`"response_format": {"type": "json_object"}`) has no effect — the backend receives a request without the JSON constraint and returns plain text.

## Steps to reproduce

Send a JSON mode request through BBR with an Anthropic ExternalModel pointing to llm-katan simulator:

```bash
curl http://<gateway>/llm/<anthropic-model>/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "<model>",
    "messages": [{"role": "user", "content": "list 3 colors as JSON"}],
    "response_format": {"type": "json_object"}
  }'
```

## Expected

Response content should be valid JSON (llm-katan echo wraps content in a JSON object when `response_format` is set).

## Actual

Response content is plain text — the `response_format` field was stripped by the translator before reaching the simulator.

## Notes

Anthropic's native API doesn't have a direct `response_format` equivalent, but it does support JSON mode via a `"type": "json"` parameter or system prompt injection. The translator could either:
- Map `response_format.type=json_object` to the Anthropic-equivalent mechanism
- Pass it through as-is for OpenAI-compatible endpoints behind Anthropic providers

## Environment

- Discovered via automated E2E tests on local Kind deployment with llm-katan v0.15.0 simulator
- Affects `pkg/plugins/api-translation/translator/anthropic/`
- Bug exists on `main` — unrelated to ExternalProvider CRD work

yossiovadia:
## Problem

The Anthropic api-translation plugin silently drops the `response_format` field when translating OpenAI → Anthropic requests. This means JSON mode (`response_format: {type: json_object}`) has no effect for Anthropic providers — the backend receives a request without the JSON constraint and returns plain text.

Fixes #186

## Fix

Pass `response_format` through to the translated request body (4 lines). Anthropic's Messages API supports this field, and backends that don't recognize it will ignore it.

## Test plan

- [x] Unit tests: `TestTranslateRequest_ResponseFormat` (passes through) + `TestTranslateRequest_NoResponseFormat` (not present when not in input)
- [x] `go test ./pkg/plugins/api-translation/translator/anthropic/` — all tests pass
- [x] E2E verified on `maas-local` (old MaaS CRDs) — Anthropic JSON mode returns valid JSON
- [x] E2E verified on `maas-experimental` (new ExternalProvider CRDs) — same result
- [x] Bug reproduced on both clusters BEFORE fix (plain text returned)

Discovered via automated E2E tests using local Kind deployment + llm-katan v0.15.0 simulator.

## Summary by CodeRabbit

* **New Features**
  * Added support for response format specification when translating API requests through the plugin layer.

* **Tests**
  * Added test coverage for response format parameter handling in request translation, including cases with and without the parameter.

### DIFF
--- a/pkg/plugins/api-translation/translator/anthropic/anthropic.go
+++ b/pkg/plugins/api-translation/translator/anthropic/anthropic.go
@@ -98,6 +98,10 @@ func (t *AnthropicTranslator) TranslateRequest(body map[string]any) (map[string]
 		translated["stream"] = stream
 	}
 
+	if responseFormat, ok := body["response_format"]; ok {
+		translated["response_format"] = responseFormat
+	}
+
 	headers := map[string]string{
 		"anthropic-version": anthropicAPIVersion,
 		"content-type":      "application/json",
--- a/pkg/plugins/api-translation/translator/anthropic/anthropic_test.go
+++ b/pkg/plugins/api-translation/translator/anthropic/anthropic_test.go
@@ -785,3 +785,33 @@ func TestTranslateRequest_NonTextContentSkipped(t *testing.T) {
 	msgs := translated["messages"].([]map[string]any)
 	assert.Equal(t, "Describe this", msgs[0]["content"])
 }
+
+func TestTranslateRequest_ResponseFormat(t *testing.T) {
+	body := map[string]any{
+		"model":    "claude-3-opus",
+		"messages": []any{map[string]any{"role": "user", "content": "list 3 colors as JSON"}},
+		"response_format": map[string]any{
+			"type": "json_object",
+		},
+	}
+
+	translated, _, _, err := NewAnthropicTranslator().TranslateRequest(body)
+	require.NoError(t, err)
+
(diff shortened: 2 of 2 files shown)


noyitz (approved):
Tested and verified on a live OpenShift cluster (sandbox2228). Confirmed the bug: sending `response_format: {type: "json_object"}` to the Anthropic translator results in markdown-wrapped JSON instead of raw JSON — the field is silently dropped during translation.

Fix is correct — same pattern as the `stream` field passthrough.

/lgtm

noyitz:
Great catch Yossi — these silent field drops are easy to miss. Nice work on the E2E verification across both cluster setups too.

PR #212, opened 2026-04-29
feat: use apiFormat for explicit API translator selection

yossiovadia:
## Problem

The api-translation plugin selects the translator based on the provider type (e.g., "openai", "anthropic") from CycleState. This is inferred from `ExternalProvider.spec.provider` and works for simple cases, but breaks when a single provider hosts models with different API formats — e.g., Bedrock serves both OpenAI-compatible and Anthropic-format models on the same endpoint.

## Fix

The api-translation plugin now prefers the `apiFormat` field (from `ExternalModel.spec.externalProviderRefs[].apiFormat`) over the provider type when selecting the translator. Falls back to provider type for backward compatibility.

```
apiFormat set → use apiFormat as translator key
apiFormat empty → fall back to provider type (existing behavior)
```

Depends on #182 (which makes apiFormat required per Nir's review).

## Changes

- `state-keys.go` — add `APIFormatKey`
- `model_store.go` — add `apiFormat` to `externalModelInfo`
- `external_model_reconciler.go` — extract `apiFormat` from unstructured ExternalModel
- `plugin.go` (model-provider-resolver) — write `apiFormat` to CycleState
- `plugin.go` (api-translation) — add `resolveTranslatorKey()` that prefers `APIFormatKey` over `ProviderKey`

## Test plan

- [x] `TestProcessRequest_APIFormatOverridesProvider` — apiFormat takes precedence over provider
- [x] `TestProcessRequest_APIFormatDirect` — apiFormat alone selects translator
- [x] `TestProcessRequest_FallsBackToProvider` — backward compatible when apiFormat not set
- [x] All existing tests pass (error message updated from "unsupported provider" to "unsupported api format")
- [x] E2E verified on maas-experimental — OpenAI, Vertex, Anthropic all returning 200


nirrozenbaum:
api format already added to the CRDs in other PRs

Issue #205, opened 2026-04-28
Nemo guardrails: support system role in /v1/guardrail/checks endpoint


liavweiss:
## Problem
The TrustyAI NeMo Guardrails server (`quay.io/trustyai/nemo-guardrails-server:latest`) rejects messages with `role: "system"` in the `/v1/guardrail/checks` endpoint, raising a `ValueError`:
```
Unsupported message role: 'system'. Supported roles are: 'user', 'assistant', 'tool'.
```
**Problematic line:**
[`server/api.py#L1041`](https://github.com/trustyai-explainability/NeMo-Guardrails/blob/6a3cfd1e51fd978075adc4355846dcea5927d60c/nemoguardrails/server/api.py#L1041) — the `_process_message` function raises `ValueError` for any role other than `user`, `assistant`, or `tool`.

## Security Impact
Because the endpoint returns an error for `system` messages, our BBR plugin (`nemo-request-guard`) is forced to filter them out before sending to NeMo. This means an attacker can inject malicious content in a `system` role message and completely bypass the guardrails.
See: [#160](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/160) and the workaround in [#162](https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/162).

## Expected Behavior
The endpoint should accept `system` role messages and run input rails on them, consistent with how the upstream NVIDIA NeMo Guardrails library handles `system` messages.
## Current Workaround
The BBR plugin filters out `system` messages before sending to NeMo.

## Action Items
1. **TrustyAI side:** Add `system` role support.
2. **BBR plugin side:** Once TrustyAI supports `system`, remove the filtering logic in `nemo-request-guard` so all messages (including `system`) are forwarded to NeMo.

cc: @christinaexyou

christinaexyou:
@liavweiss happy to work on this - can I be assigned to this issue ?

nirrozenbaum:
/assign @christinaexyou 

Thanks!!

liavweiss:
Thanks @christinaexyou, are you also working on the plugin side? I’m happy to take it on if needed.

nirrozenbaum:
looks like https://github.com/trustyai-explainability/NeMo-Guardrails/pull/47 was merged, system role should be valid now.

PR #232, opened 2026-05-04
fix: remove system role filtering, TrustyAI now supports system messages

### LINKED ISSUE #160: security issue in nemo guard plugin

liavweiss:
## Summary

- Remove system message filtering from `nemo-request-guard`. Previously, messages with `role: "system"` were filtered out before sending to NeMo because the TrustyAI `/v1/guardrail/checks` endpoint rejected them with a hard error (`Unsupported message role: 'system'`).
- TrustyAI has since merged [feat(api): Support input guardrail checks when role: 'system' (#47)](https://github.com/trustyai-explainability/NeMo-Guardrails/pull/47).
- All messages (including system) are now forwarded to NeMo for full conversation context evaluation.

Closes #160. #205 

## Changes

- `request_guard.go`: Removed the `if role == "system" { continue }` filter in `extractOpenAIMessages`. Updated comments.
- `request_guard_test.go`: Updated tests to expect system messages in the forwarded payload.

## E2E verification
Verified end-to-end by building the TrustyAI NeMo image locally from the `develop` branch (`Dockerfile.server`) which includes PR #47, and deploying it on a Kind cluster with `detect sensitive data on input` rails:
| Test | Messages | Result |
|------|----------|--------|
| Safe system + user | `system: "You are a helpful assistant"` + `user: "Hello"` | **success** (both checked) |
| PII in system msg | `system: "card 4111-1111-1111-1111"` + `user: "Hello"` | **blocked** (system detected) |
| System-only with PII | `system: "Contact John Smith at john@example.com"` | **blocked** (system detected) |

cc @nirrozenbaum @christinaexyou 

## Summary by CodeRabbit

* **Bug Fixes**
  * System messages are now included when forwarding conversations to NeMo, ensuring complete message history is transmitted.
* **Tests**
  * Unit tests updated to expect and validate all message roles (including system) and adjusted cases for single-system vs. system-only conversations.

### DIFF
--- a/pkg/plugins/nemo/request_guard.go
+++ b/pkg/plugins/nemo/request_guard.go
@@ -133,7 +133,7 @@ func (p *NemoRequestGuardPlugin) ProcessRequest(ctx context.Context, _ *framewor
 // extractMessages returns user-supplied text as a message slice suitable for NeMo's
 // OpenAI-compatible chat endpoint. It supports two payload formats:
 //
-//  1. OpenAI chat: top-level "messages" array → forwards all non-system messages.
+//  1. OpenAI chat: top-level "messages" array → forwards all messages.
 //  2. MCP JSON-RPC: {"jsonrpc":"2.0","params":{"arguments":{…}}} → concatenates
 //     all string argument values into a single user message.
 //
@@ -148,15 +148,17 @@ func extractMessages(body map[string]any) ([]map[string]string, error) {
 	return nil, nil // not an inference request (e.g. API key management, model listing)
 }
 
-// extractOpenAIMessages parses an OpenAI-style "messages" value. System messages are
-// filtered. TrustyAI does not support this role. All other roles are forwarded so NeMo can evaluate
-// the full conversation context.
+// extractOpenAIMessages parses an OpenAI-style "messages" value. All messages are forwarded
+// so NeMo can evaluate the full conversation context.
 func extractOpenAIMessages(raw any) ([]map[string]string, error) {
 	slice, ok := raw.([]any)
 	if !ok {
 		return nil, fmt.Errorf("messages is not an array")
 	}
-	var messages []map[string]string
+	if len(slice) == 0 {
+		return nil, nil
+	}
+	messages := make([]map[string]string, 0, len(slice))
 
 	for _, m := range slice {
 		msg, ok := m.(map[string]any)
@@ -164,9 +166,6 @@ func extractOpenAIMessages(raw any) ([]map[string]string, error) {
 			continue
 		}
 		role, _ := msg["role"].(string)
-		if role == "system" {
-			continue
-		}
 		content, _ := msg["content"].(string)
 		messages = append(messages, map[string]string{"role": role, "content": content})
 	}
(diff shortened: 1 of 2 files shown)


nirrozenbaum:
@liavweiss making sure - was this change tested with latest trusty version?

pkg/plugins/nemo/request_guard.go
```diff
@@ -164,9 +163,6 @@ func extractOpenAIMessages(raw any) ([]map[string]string, error) {
 			continue
 		}
 		role, _ := msg["role"].(string)
-		if role == "system" {
```
nirrozenbaum, line 167: would it be correct to just return “slice” without copying?
in current code we copy the whole messages array on every request. 
so if we have a long multi turn chat, memory allocation for each turn is expensive.

liavweiss:
> @liavweiss making sure - was this change tested with latest trusty version?

Yes, I built the TrustyAI NeMo image locally that includes the new code https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/47

pkg/plugins/nemo/request_guard.go
liavweiss: Thanks, I added a pre-allocation and an early return for empty slices to avoid unnecessary allocations.

nirrozenbaum: my question was actually about the messages themselves, not only about the case of empty slice.
my intention was - if we have something along these lines:
```go
func extractOpenAIMessages(raw any) ([]map[string]string, error) {
	slice, ok := raw.([]any)
	if !ok {
		return nil, fmt.Errorf("messages is not an array")
	}

    if len(slice) == 0 {
		return nil, nil
	}

    return slice
```

would it work?

liavweiss: Should work with some extra changes, but returning the slice directly means forwarding all original message fields to NeMo (e.g., `tool_calls`, f`unction_call`) instead of only `role` and `content`. NeMo ignores extra fields, but it sends more data over the wire for no reason. It also requires changing the return type from `[]map[string]string` to `[]any` and updating all callers and tests accordingly. If we want to go in that direction, it would be better as a follow-up PR.

nirrozenbaum: sure. I'm just thinking about the tradeoffs. 
do we prefer allocating memory for every request and copying the whole chat? (that could be very large) or do we prefer sending the same request that we received over the wire to yet another http hop? (but at least we don't do both - allocate memory + sending http). 

there are pros/cons for both sides. 
my gut feeling says sending the original request as is would probably be "cheaper". 
we can handle in a follow up, for sure.

nirrozenbaum (approved):
/lgtm
/approve

PR #243, opened 2026-05-07
fix(security): enforce namespace boundary for external model resolution

henschwartz:
## Summary

Adds **namespace boundary enforcement** in `model-provider-resolver` before resolving
`ExternalModel` and wiring credential references downstream. This closes a
tenant-isolation gap where a client could target another namespace’s model path
to trigger use of secrets in that namespace.

Fixes #242

## Root cause

Request path parsing derives `namespace/name` from `:path` (`/<ns>/<model>/v1/...`)
without verifying the **requester’s authorized namespace** matches the target
model namespace.

## Fix

- Resolve requester namespace from (in order):
  - `x-ai-gateway-request-namespace`
  - `x-kubernetes-namespace`
  - `x-namespace`
  - `x-forwarded-client-cert` SPIFFE URI (`.../ns/<namespace>/...`)
- If requester namespace cannot be determined → **`403 Forbidden`**
- If requester namespace differs from model namespace parsed from path → **`403 Forbidden`**
- Same-namespace requests continue through the existing plugin chain.

## Tests

- Updated/added unit coverage in:
  - `pkg/plugins/model-provider-resolver/plugin_test.go`
- Updated e2e request builder to send:
  - `x-ai-gateway-request-namespace: <test namespace>`
  in `test/e2e/e2e_test.go` so existing e2e curls match the new policy.

## Manual verification (local)

Validated on kind/istio deploy that:
- cross-namespace request is blocked with `403`
- missing requester namespace context is blocked with `403`

## Notes / follow-ups

- This assumes the mesh/gateway injects a trusted requester namespace signal
  (header or XFCC). If deployments need a different canonical header, we can
  extend the extractor list in one place.

## Release Notes

* **Bug Fixes**
  * Added namespace-boundary validation for external model requests. The system now verifies that request namespaces match the external model's namespace configuration. Requests with namespace mismatches are rejected to prevent unauthorized access.

### DIFF
--- a/pkg/plugins/model-provider-resolver/plugin.go
+++ b/pkg/plugins/model-provider-resolver/plugin.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


nirrozenbaum:
please see this comment: 
https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/242#issuecomment-4396107252

/hold

henschwartz:
Closing this based on previous comment and a second look at the actual threat model.
Payload processing runs after MaaS (auth + rate limiting), and the CRD already constrains ExternalModel / provider / secretRef to the same namespace.
We’re not seeing a concrete residual gap here anymore, so we’re closing the ticket (and the related PR)

PR #249, opened 2026-05-07
fix: validate required CRD fields in ExternalModel reconciler

liavweiss:
## Summary

- Validate that required fields (`spec.provider`, `spec.targetModel`, `spec.credentialRef.name`) extracted from ExternalModel CRD are non-empty before storing them in the model info store
- Previously, `unstructured.NestedString` return values (`found`, `error`) were silently discarded with `_, _`, allowing empty/invalid data to be stored
- On validation failure, stale store entries are removed before returning the error to prevent serving incomplete data
- Returning the error causes controller-runtime to requeue with backoff, surfacing the misconfiguration in controller logs

## Test
- `go build ./...` passes

## Related
#close https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/248

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation of external model configurations. Required fields are now verified before processing. Invalid or incomplete configurations are automatically removed with clear error messages identifying which fields are missing or invalid.

### DIFF
--- a/pkg/plugins/model-provider-resolver/external_model_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_model_reconciler.go
@@ -62,9 +62,23 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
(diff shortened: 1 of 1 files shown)


nirrozenbaum:
the CRDs are enforcing strings with min length of at least 1 char for mandatory fields.. so this PR is unnecessary. 
example:
https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/5741b70f35026e0ccfc2a0a901340c41eea7247f/api/inference/v1alpha1/externalprovider_types.go#L45

PR #251, opened 2026-05-07
fix(security): harden sanitizePath against path traversal and encodin…

mkoushni:
## Summary

Fixes insufficient path sanitization in the `model-provider-resolver` plugin (`pkg/plugins/model-provider-resolver/plugin.go`).

The original `sanitizePath` function only trimmed whitespace and stripped query parameters, leaving multiple attack vectors unaddressed.

## Vulnerability Details

**Location:** `pkg/plugins/model-provider-resolver/plugin.go:140-147`

The following protections were missing:

| Attack Vector | Example | Fix Applied |
|---|---|---|
| Path traversal | `/../etc/passwd`, `../../secret` | Reject segments equal to `..` on raw decoded path |
| Windows backslash traversal | `..\etc\passwd` | Split on `\` within each segment and reject `..` |
| URL encoding bypass | `%2e%2e`, `%2f`, `%2E%2E` | URL-decode with `url.PathUnescape` before traversal check |
| Null byte injection | `\x00`, `%00` | Reject before and after URL decoding |
| Control character injection | `\n`, `\r`, `0x01`–`0x1F`, DEL `0x7F` | Scan all runes before processing |
| Fragment injection | `path#fragment` | Strip `#` and everything after it |
| Invalid percent-encoding | `%GG` | Return error on `url.PathUnescape` failure |

### `plugin.go`
- Added `net/url` and `path` imports
- `sanitizePath` now returns `(string, error)` instead of `string`
- Guards are applied in a deliberate order:
  1. Null byte check (raw)
  2. Control character scan
  3. Fragment stripping (`#`)
  4. Query parameter stripping (`?`)
  5. URL decoding + post-decode null byte check
  6. Path traversal check on **raw decoded segments** (before `path.Clean`)
  7. Final normalization via `path.Clean`
- `ProcessRequest` handles the new error and returns `BadRequest` for invalid paths

> **Note:** Traversal detection runs on the decoded-but-not-yet-cleaned path intentionally — checking after `path.Clean` would be incorrect because `Clean` silently resolves `..` segments, making detection impossible.

### `plugin_test.go`
- Added `TestSanitizePath` with 25 table-driven sub-tests covering all attack classes and happy paths

## Test Coverage
=== RUN TestSanitizePath
--- PASS: normal_path 
--- PASS: path_with_query_params_stripped 
--- PASS: path_with_fragment_stripped 
--- PASS: path_with_both_query_and_fragment_stripped 
--- PASS: leading_and_trailing_slashes_trimmed 
--- PASS: surrounding_whitespace_trimmed 
--- PASS: path_traversal_with_dot-dot-slash 
--- PASS: path_traversal_at_root 
--- PASS: path_traversal_with_backslash_notation 
--- PASS: path_traversal_multiple_levels 
--- PASS: URL-encoded_dot-dot_(%2e%2e)traversal 
--- PASS: URL-encoded_slash(%2f)in_traversal 
--- PASS: mixed_case_URL-encoded_traversal(%2E%2E) 
--- PASS: null_byte_in_path 
--- PASS: percent-encoded_null_byte_(%00) 
--- PASS: control_character_0x01_in_path 
--- PASS: control_character_DEL_(0x7F)_in_path 
--- PASS: newline_character_in_path 
--- PASS: carriage_return_in_path 
--- PASS: invalid_percent-encoding 
--- PASS: TestSanitizePath

## Related
Closes: [RHOAIENG-59380](https://redhat.atlassian.net/browse/RHOAIENG-59380)

### DIFF
--- a/pkg/plugins/model-provider-resolver/plugin.go
+++ b/pkg/plugins/model-provider-resolver/plugin.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"path"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -102,7 +104,10 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 	}
 
 	log.FromContext(ctx).V(logutil.VERBOSE).Info("received incoming request", "path", request.Headers[":path"])
-	relativePath := sanitizePath(request.Headers[":path"])
+	relativePath, err := sanitizePath(request.Headers[":path"])
+	if err != nil {
+		return errcommon.Error{Code: errcommon.BadRequest, Msg: fmt.Sprintf("invalid request path: %v", err)}
+	}
 
 	segments := strings.Split(relativePath, "/")
 	if len(segments) < 2 || segments[0] == "" || segments[1] == "" {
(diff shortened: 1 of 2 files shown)


nirrozenbaum (commented):
@mkoushni Thanks for the PR! 
we're actually switching to read model directly from the body and not from url (that was a limitation in model-as-a-service component), and that change would make this PR irrelevant...
payload processing shouldn't care at all about the url (that was a temp solution for maas limitation)

PR #255, opened 2026-05-10
Partition secret and model stores by namespace

abdallahsamabd:
## Description

This change addresses multi-tenant isolation concerns for in-memory caches used by the **apikey-injection** and **model-provider-resolver** plugins ([issue #250](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/250)).

**`secretStore` (`pkg/plugins/apikey-injection`)**  
Previously, credentials were held in a single flat `map[string]…` keyed by a composite `"namespace/name"` string. The store now uses a **two-level map**: `namespace → secret name → credential fields`. Public behavior is unchanged for reconciler and request flow, but reads/writes/deletes use **explicit `(namespace, name)`** so the data layout matches Kubernetes namespacing and avoids a single global string keyspace. Empty namespace buckets are removed when the last secret in that namespace is deleted.

**`modelInfoStore` (`pkg/plugins/model-provider-resolver`)**  
External model metadata was stored in a flat `map[string]*externalModelInfo` keyed by `NamespacedName.String()`. The store now partitions internally as **`namespace → external model name → info`**, while keeping the same method signatures that take `types.NamespacedName`. Deletes clear the inner entry and drop the namespace map when it becomes empty.

**Tests**  
Unit tests were updated for the new `secretStore` API. Cross-namespace isolation cases were added for both stores so a row in namespace A is not visible when querying with namespace B and the same resource name.

**Note:** This hardens cache structure and API clarity; **authorization** (who may choose which namespace on the request path) remains governed by deployment, routing, and Kubernetes RBAC outside these stores.

## How Has This Been Tested?

- **Local / CI-style unit tests:**  
  `go test ./pkg/plugins/apikey-injection ./pkg/plugins/model-provider-resolver`  
  (all tests in those packages passed.)

- **Build:**  
  `go build ./...`  
  (succeeded.)

- **Full `go test ./...`:**  
  Fails in `test/e2e` when no Kubernetes API server is available at `127.0.0.1:6443` (expected in environments without a cluster); plugin unit tests were run explicitly as above.

- **Environment:** Windows, Go toolchain as used by the repo (no e2e cluster in this run).

## Merge criteria:

## Summary by CodeRabbit

* **Refactor**
  * Reworked internal stores to use explicit namespace + name addressing, improving namespace isolation and lifecycle handling.

* **Tests**
  * Updated and added tests to validate namespace-level isolation, concurrent access, and correct add/update/delete behavior for credentials and model info.

### DIFF
--- a/pkg/plugins/apikey-injection/store.go
+++ b/pkg/plugins/apikey-injection/store.go
@@ -27,30 +27,32 @@ import (
 // namespaced name ("namespace/name") to its credential data.
 // The secretReconciler writes to it; the apiKeyInjectionPlugin reads from it.
 type secretStore struct {
-	mu   sync.RWMutex
-	data map[string]map[string]string
+	mu sync.RWMutex
+	// namespace -> secret name -> credentials
+	data map[string]map[string]map[string]string
 }
 
 // newSecretStore creates an empty secretStore.
 func newSecretStore() *secretStore {
 	return &secretStore{
-		data: make(map[string]map[string]string),
+		data: make(map[string]map[string]map[string]string),
 	}
 }
 
 // addOrUpdate extracts all fields from the Secret's data and stores them under
 // the given key. Returns an error if the Secret has no data fields or if any
 // field is empty.
-func (s *secretStore) addOrUpdate(key string, secret *corev1.Secret) error {
+func (s *secretStore) addOrUpdate(namespace, name string, secret *corev1.Secret) error {
+	key := fmt.Sprintf("%s/%s", namespace, name)
 	if len(secret.Data) == 0 {
-		s.delete(key)
+		s.delete(namespace, name)
 		return fmt.Errorf("secret '%s' has no data fields", key)
 	}
 
 	credentials := make(map[string]string)
 	for field, value := range secret.Data {
 		if len(value) == 0 {
-			s.delete(key)
+			s.delete(namespace, name)
 			return fmt.Errorf("secret '%s' has empty field '%s'", key, field)
 		}
 		credentials[field] = string(value)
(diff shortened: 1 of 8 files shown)


pkg/plugins/apikey-injection/store.go
```diff
@@ -27,52 +27,70 @@ import (
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	delete(s.data, secretKey)
+	modelsByNamespace, found := s.data[namespace]
+	if !found {
+		return
+	}
+	delete(modelsByNamespace, name)
+	if len(modelsByNamespace) == 0 {
+		delete(s.data, namespace)
```
senanz: **Must fix — copy-paste naming from the model store.**

`modelsByNamespace` is misleading here: this is the *secret* store, not the model store. The variable holds a map of secret names to credentials, so it should be named `secretsByName` (or `namespacedSecrets`).

CodeRabbit caught this in `delete`, but the same issue also appears below in `get` (line ~87). Both sites need renaming in the same pass.

```suggestion
	secretsByName, found := s.data[namespace]
	if !found {
		return
	}
	delete(secretsByName, name)
	if len(secretsByName) == 0 {
		delete(s.data, namespace)
	}
```

pkg/plugins/apikey-injection/store.go
```diff
@@ -27,52 +27,70 @@ import (
...
-func (s *secretStore) get(secretKey string) (map[string]string, bool) {
+func (s *secretStore) get(namespace, name string) (map[string]string, bool) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
-	credentials, ok := s.data[secretKey]
+
+	modelsByNamespace, found := s.data[namespace]
+	if !found {
+		return nil, false
+	}
```
senanz, line 92: Same naming issue here — `modelsByNamespace` should be `secretsByName` to match the domain of this store.

```suggestion
	secretsByName, found := s.data[namespace]
	if !found {
		return nil, false
	}

	credentials, ok := secretsByName[name]
	return credentials, ok
```

senanz (changes requested):
Good, well-scoped refactoring — the namespace partitioning logic is correct and the cross-namespace isolation tests cover the critical invariant. One issue that needs fixing before merge:

## Must fix

**`modelsByNamespace` variable name in `store.go` (`delete` and `get`)**

This is a clear copy-paste artifact from the parallel model store work. The secret store's `delete` and `get` methods both declare a local variable `modelsByNamespace` to hold the inner map, but this store holds *secrets*, not models. Rename to `secretsByName` in both methods. I've left inline suggestions on both sites.

This is the kind of thing that will confuse the next engineer who reads a stack trace or steps through a debugger in this code — the variable name contradicts the surrounding context.

---

### Note on CodeRabbit's findings

CodeRabbit raised this same variable naming issue but only covered the `delete` method. The `get` method has the same problem. The other two CodeRabbit comments (pointer aliasing / nil guard on `modelInfoStore`) are valid defensive-programming observations, but they apply equally to the pre-existing code on `main` and are not regressions introduced by this PR. I'd suggest addressing those in a separate hardening pass rather than expanding the scope of this change.

pkg/plugins/apikey-injection/store.go
abdallahsamabd: done

pkg/plugins/apikey-injection/store.go
abdallahsamabd: done

szedan-rh:
LGTM

szedan-rh:
@nirrozenbaum - can you also do a 2nd eye review?

szedan-rh (approved):

Issue #123, opened 2026-04-14
Add redaction (masking) support to NeMo guard plugins


liavweiss:
The `nemo-request-guard` and `nemo-response-guard` plugins currently only support two outcomes from NeMo: **allow** (`status: success`) or **block** (`status: blocked` → HTTP 403).
NeMo Guardrails also supports a third outcome: **modified** — where sensitive data (PII, financial, etc.) is masked in the message (e.g. `"My name is {PERSON}"`) and the request/response should continue with the redacted content.
The plugins should be extended to handle this third status:
- Read the `messages` field from NeMo's response
- If the content was modified (redacted), replace the original request/response body with the redacted version before forwarding
- This applies to both `nemo-request-guard` (input redaction) and `nemo-response-guard` (output redaction)
### Clarification
- The plugin does **not** perform any redaction logic itself — NeMo handles all detection and masking based on its configuration.
- The plugin's only new responsibility is to detect when NeMo has modified the content and apply the rewritten messages to the request/response body
- What entities to redact (PII, credit cards, custom patterns) is entirely a NeMo configuration concern, not a plugin concern
- The NeMo pod could be deployed manually, through TrustyAI, or any other way — the plugin just calls the endpoint
### Dependencies
- The `nemo-response-guard` plugin is currently in review: [#113](https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/113)
- The shared `nemo_guard.go` (introduced in #113) contains the `callNemoGuard()` function that will need to be updated to return the redacted messages when applicable
- This issue should be worked on after #113 is merged, as the redaction changes build on top of the shared guard infrastructure and the response guard
### Acceptance criteria
- [ ] When NeMo returns a modified/redacted message, the plugin replaces the original body and forwards to the model (request guard) or back to the caller (response guard)
- [ ] When NeMo returns `success` (no modification), behavior is unchanged — forward as-is
- [ ] When NeMo returns `blocked`, behavior is unchanged — HTTP 403
- [ ] Unit tests cover all three paths: allow, block, redact
- [ ] The shared `callNemoGuard` logic in `nemo_guard.go` is updated to return the redacted messages when applicable

nirrozenbaum:
@liavweiss thanks for creating the issue. 
do you know if nemo/trusty returns any indication if the content was redacted or not? 
otherwise, we would need to do deep equal check (which could be very expensive operation in long response).

[...]

PR #264, opened 2026-05-12
feat(nemo): migrate guard plugins from /v1/guardrail/checks endpoint to /v1/chat/completions endpoint

liavweiss:
## Summary

- Refactor `nemo-request-guard` and `nemo-response-guard` plugins to use the NeMo `/v1/chat/completions` endpoint instead of the legacy `/v1/guardrail/checks` endpoint
- Add `message_action` / `block_reason` support via NeMo's `guardrails.options.output_vars` mechanism to distinguish pass, block, and modify (redact) actions from the response
- Introduce configurable `actionVar` and `reasonVar` fields in plugin config (defaulting to `message_action` and `block_reason`) so admins can match their NeMo Colang variable names
- Replace the old `nemoResponse` / `nemoRailStatusEntry` types with new `nemoCompletionResponse` / `nemoGuardResult` types for the chat-completions response format
- Consolidate request marshaling into `callNemoGuard` - callers now pass `model` + `messages` instead of pre-marshaled bytes
- Return `errcommon.Error` directly from `callNemoGuard` with specific codes (`Internal` for marshaling errors, `ServiceUnavailable` for network/response errors) instead of separate error code + error returns
- Forward NeMo's `block_reason` to the client in the 403 response instead of a generic message
- Treat `modify` action as pass-through for now - full redaction (message rewriting) will be added as part of https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/123

## Tested

- All existing unit tests updated to use the new `/v1/chat/completions` response format and pass
- New test cases added for `modify` action (pass-through) and block-without-reason scenarios
- `go build ./...` and `go test ./pkg/plugins/nemo/` pass with no errors or lint issues
- E2E tested on a Kind cluster with a NeMo Guardrails deployment using Rob's Colang config that sets `message_action` and `block_reason`:
  - Normal message - plugin returns `action: pass`, request forwarded to backend (HTTP 200)
  - Forbidden words ("I hate everyone") - plugin returns `action: block`, client gets HTTP 403 with `"Forbidden words found in input"`
  - PII message ("My name is John Smith") - plugin returns `action: modify`, request passes through (redaction is follow-up PR)

## Context

- TrustyAI will maintain a catalog of standardized Colang configs that set `message_action` to `"pass"`, `"block"`, or `"modify"` and populate `block_reason`
- The legacy `/v1/guardrail/checks` endpoint only supported blocking; the new endpoint enables both blocking and PII redaction
- Follow-up PRs: redaction support (rewrite message content on `modify`), unit tests for `nemo_guard_base.go`, documentation update

## Related
#256 

## Summary by CodeRabbit

* **Refactor**
  * NeMo guardrails updated to a chat-style request/response flow; decisions now return structured actions (pass/modify/block) with optional reasons and content, and block responses surface a clear reason.

* **Tests**
  * Guardrail test suites and payload assertions updated to validate the new action/reason response format and decision semantics.

### DIFF
--- a/pkg/plugins/nemo/nemo_guard_base.go
+++ b/pkg/plugins/nemo/nemo_guard_base.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -33,34 +32,61 @@ import (
 )
 
 const (
-	nemoAllowedStatus    = "success"
 	defaultTimeoutSec    = 360
 	maxNemoResponseBytes = 1 << 20 // 1 MiB
+
+	defaultActionVar = "message_action"
+	defaultReasonVar = "block_reason"
+
+	ActionPass   = "pass"
+	ActionBlock  = "block"
+	ActionModify = "modify"
 )
 
 // nemoGuardConfig is the configuration for nemo guard plugins.
 type nemoGuardConfig struct {
 	NemoURL        string `json:"nemoURL"`
 	TimeoutSeconds int    `json:"timeoutSeconds"`
+	ActionVar      string `json:"actionVar"`
+	ReasonVar      string `json:"reasonVar"`
 }
 
-// nemoResponse is NeMo's JSON response from /v1/guardrail/checks.
-type nemoResponse struct {
-	Status      string                         `json:"status"`
-	RailsStatus map[string]nemoRailStatusEntry `json:"rails_status"`
+// nemoCompletionResponse is NeMo's JSON response from /v1/chat/completions.
+type nemoCompletionResponse struct {
+	Choices    []nemoChoice       `json:"choices"`
+	Guardrails nemoGuardrailsData `json:"guardrails"`
 }
 
-type nemoRailStatusEntry struct {
-	Status string `json:"status"`
+type nemoChoice struct {
+	Message nemoMessage `json:"message"`
+}
+
+type nemoMessage struct {
+	Content string `json:"content"`
+	Role    string `json:"role"`
+}
+
+type nemoGuardrailsData struct {
+	ConfigID   string         `json:"config_id"`
+	OutputData map[string]any `json:"output_data"`
+}
+
+// nemoGuardResult holds the parsed outcome from a /v1/chat/completions call.
+type nemoGuardResult struct {
+	Action  string // ActionPass, ActionBlock, or ActionModify
+	Reason  string // human-readable reason (empty for pass)
+	Content string // response content from choices[0].message.content
 }
 
 // nemoGuardBase holds the shared fields and HTTP logic for nemo guard plugins.
 type nemoGuardBase struct {
 	nemoURL    string
 	httpClient *http.Client
+	actionVar  string
+	reasonVar  string
 }
 
-func newNemoGuardBase(nemoURL string, timeoutSeconds int) (*nemoGuardBase, error) {
+func newNemoGuardBase(nemoURL string, timeoutSeconds int, actionVar, reasonVar string) (*nemoGuardBase, error) {
 	if nemoURL == "" {
 		return nil, errors.New("nemoURL is required")
 	}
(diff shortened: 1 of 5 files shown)


pkg/plugins/nemo/nemo_guard_base.go
```diff
@@ -33,101 +32,153 @@ import (
...
 // nemoGuardConfig is the configuration for nemo guard plugins.
 type nemoGuardConfig struct {
 	NemoURL        string `json:"nemoURL"`
 	TimeoutSeconds int    `json:"timeoutSeconds"`
+	ActionVar      string `json:"actionVar"`
+	ReasonVar      string `json:"reasonVar"`
+}
+
+// nemoCompletionResponse is NeMo's JSON response from /v1/chat/completions.
+type nemoCompletionResponse struct {
```
nirrozenbaum, line 55: this could be confusing for a reader because there is a `completion` (deprecated) api.
probably better to name it either `nemoChatCompleionResponse` or just `nemoResponse` under the assumption that it's always chat completion (which the current code assumes)

pkg/plugins/nemo/request_guard.go
```diff
@@ -89,12 +90,13 @@ func (p *NemoRequestGuardPlugin) WithName(name string) *NemoRequestGuardPlugin {
 
 // ProcessRequest calls NeMo Guardrails to evaluate input rails on the incoming request.
 // It extracts user-supplied text from either an OpenAI-style chat body (via "messages")
-// or an MCP JSON-RPC body (via "params.arguments"), POSTs to NeMo url, and returns an
-// errcommon.Error with Forbidden (403) if NeMo flags the content.
+// or an MCP JSON-RPC body (via "params.arguments"), POSTs to NeMo's /v1/chat/completions
```
nirrozenbaum, line 93: super nit: no need to specify the relative url.. this is configurable.
```suggestion
// or an MCP JSON-RPC body (via "params.arguments"), POSTs to NeMo's 
```

pkg/plugins/nemo/request_guard.go
```diff
@@ -109,25 +111,26 @@ func (p *NemoRequestGuardPlugin) ProcessRequest(ctx context.Context, _ *framewor
...
-		if code == errcommon.Forbidden {
-			return errcommon.Error{Code: code, Msg: "request blocked by NeMo guardrails"}
+		return err
+	}
+
+	switch result.Action {
+	case ActionPass:
+		return nil
+	case ActionModify:
+		// TODO: support redaction by rewriting the message content.
```
nirrozenbaum, line 123: do we get back the redacted content in chat completions format?
I mean, is it a drop in instead of the existing body and that's it? or should we do here any special processing?

pkg/plugins/nemo/nemo_guard_base.go
```diff
@@ -33,101 +32,153 @@ import (
...
 	if err := json.Unmarshal(body, &nemoResp); err != nil {
 		logger.Error(err, "failed to decode NeMo response")
-		return errcommon.ServiceUnavailable, fmt.Errorf("failed to decode nemo response: %w", err)
+		return nil, errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: fmt.Sprintf("failed to decode nemo response: %v", err)}
 	}
 
-	if strings.EqualFold(strings.TrimSpace(nemoResp.Status), nemoAllowedStatus) {
-		logger.Info("allowed by NeMo guardrails")
-		return "", nil
+	result := &nemoGuardResult{Action: ActionPass}
```
nirrozenbaum, line 167: why do we default to `ActionPass` and not just empty struct (which could provider `unknown` in the upper layer)?

liavweiss:
Putting this PR on hold pending an API change from TrustyAI. see [comment](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/256#issuecomment-4477063383)

/hold

liavweiss:
/close
See https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/256#issuecomment-4489027160

PR #269, opened 2026-05-13
feat: add provider store and cross-watch to model-provider-resolver

yossiovadia:
## Summary

Split from PR #184 per Nir's request for smaller PRs.

- Add ExternalProvider store reconciler (unstructured, read-only) that caches provider endpoint, credentials, and config
- Update ExternalModel store reconciler to cross-reference the provider store for credential/config resolution
- Cross-watch: provider changes trigger re-reconciliation of dependent ExternalModels
- Propagate provider config via CycleState (`ProviderConfigKey`) for downstream plugins

This is the data-plane half — store reconcilers that read CRDs and populate CycleState. The control-plane half (typed controller reconcilers + RBAC) follows in a separate PR.

## Files changed (9, +711/-40)

| File | What |
|------|------|
| `external_provider_reconciler.go` + test | New: watches ExternalProvider CRDs, populates provider store |
| `provider_store.go` + test | New: thread-safe store for provider info (endpoint, creds, config) |
| `external_model_reconciler.go` + test | Updated: reads provider store for credential/config resolution |
| `model_store.go` | Add config field |
| `state-keys.go` | Add `ProviderConfigKey` |
| `plugin.go` | Register provider store reconciler + cross-watch |

## Test plan

- [x] All unit tests pass (`make test-unit`)
- [x] Tested on Kind cluster (maas-experimental): provider store populates correctly, config propagates through CycleState
- [x] 32 e2e-integration tests pass (reconciler, auth, lifecycle)

cc @nirrozenbaum

## Summary by CodeRabbit

* **New Features**
  * Support for ExternalProvider resources and provider info store; models can reference external providers and provider config is persisted with model state when present.
  * Plugin now watches providers and enqueues affected models on provider changes.

* **Bug Fixes**
  * Models removed from store on deletion or missing refs; requeue when referenced providers are unavailable.
  * Stricter request handling and clearer NotFound/BadRequest responses.

* **Tests**
  * Added reconciliation and concurrency tests for provider/model flows.

* **Documentation**
  * Added controller README describing reconciliation and networking resources.


pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,36 +19,32 @@ package model_provider_resolver
...
-	store *modelInfoStore
+	modelStore    *modelInfoStore
+	providerStore *providerInfoStore
 }
 
-// Reconcile handles create/update/delete events for ExternalModel resources.
-// The ExternalModel CR name is used as the model key in the store, matching
-// the model name in inference request bodies.
 func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx).V(logutil.DEFAULT)
```
nirrozenbaum, line 50: there are few logging changes that were made and this PR reverts. 
need to put them back, like current line (log with defaul log level)

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,36 +19,32 @@ package model_provider_resolver
...
-	logger.Info("reconciling ExternalModel", "name", req.Name, "namespace", req.Namespace)
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling ExternalModel", "name", req.Name, "namespace", req.Namespace)
```
nirrozenbaum: nit: log lines should be lowercase

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	if errors.IsNotFound(err) || !obj.GetDeletionTimestamp().IsZero() {
-		r.store.deleteExternalModel(req.NamespacedName)
-		logger.Info("ExternalModel removed from store", "name", req.Name, "namespace", req.Namespace)
```
nirrozenbaum, line 63: bring back this line

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
+	refs, _, _ := unstructured.NestedSlice(obj.Object, "spec", "externalProviderRefs")
+	if len(refs) == 0 {
+		r.modelStore.deleteExternalModel(req.NamespacedName)
+		return ctrl.Result{}, nil
+	}
```
nirrozenbaum: this is not possible due to CRD validations. please remove unreachable code

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+	refMap, ok := refs[0].(map[string]any)
+	if !ok {
+		r.modelStore.deleteExternalModel(req.NamespacedName)
+		return ctrl.Result{}, nil
+	}
```
nirrozenbaum: this is also unreachable due to CRD validations

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
+	refs, _, _ := unstructured.NestedSlice(obj.Object, "spec", "externalProviderRefs")
+	if len(refs) == 0 {
+		r.modelStore.deleteExternalModel(req.NamespacedName)
+		return ctrl.Result{}, nil
+	}
+
+	refMap, ok := refs[0].(map[string]any)
```
nirrozenbaum: do we already have an open issue to revisit the code to extend single provider ref to multiple?
if not please open.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	}
+
+	providerRefName := nestedString(refMap, "ref", "name")
+	targetModel := nestedString(refMap, "targetModel")
+
+	if providerRefName == "" {
+		r.modelStore.deleteExternalModel(req.NamespacedName)
+		logger.Info("ExternalModel missing provider ref name, skipping")
+		return ctrl.Result{}, nil
+	}
```
nirrozenbaum: unreachable code, validated in CRD level. please remove.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+	providerKey := types.NamespacedName{Namespace: req.Namespace, Name: providerRefName}
+	provInfo, found := r.providerStore.get(providerKey)
+	if !found {
+		logger.Info("ExternalProvider not yet available, requeuing", "provider", providerRefName)
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
```
nirrozenbaum: why do we have this magic number 2 seconds?

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -59,24 +55,63 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	targetModel := nestedString(refMap, "targetModel")
+
+	if providerRefName == "" {
+		r.modelStore.deleteExternalModel(req.NamespacedName)
+		logger.Info("ExternalModel missing provider ref name, skipping")
+		return ctrl.Result{}, nil
+	}
+
+	providerKey := types.NamespacedName{Namespace: req.Namespace, Name: providerRefName}
+	provInfo, found := r.providerStore.get(providerKey)
```
nirrozenbaum: can we use full descriptive name? 
e.g., rename provInfo -> providerInfo

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -44,32 +45,83 @@ var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
...
-		store:  modelInfoStore,
+func NewModelProviderResolver(handle framework.Handle) (*ModelProviderResolverPlugin, error) {
+	reconcilerBuilder := handle.ReconcilerBuilder
+	k8sClient := handle.Client()
+	providerStore := newProviderInfoStore()
+	modelStore := newModelInfoStore()
+
+	// Watch ExternalProvider CRDs (inference.opendatahub.io/v1alpha1)
+	providerObj := &unstructured.Unstructured{}
+	providerObj.SetGroupVersionKind(externalProviderGVK)
```
nirrozenbaum: use typed clients now that the CRD is in this repo

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -44,32 +45,83 @@ var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
...
 
-	// Watch ExternalModel CRDs directly (no MaaS dependency)
-	obj := &unstructured.Unstructured{}
-	obj.SetGroupVersionKind(externalModelGVK)
-
-	if err := reconcilerBuilder().For(obj).Complete(reconciler); err != nil {
+	// Watch ExternalModel CRDs (inference.opendatahub.io/v1alpha1)
+	// Cross-watch ExternalProviders so credential/endpoint changes propagate to modelStore
+	modelObj := &unstructured.Unstructured{}
+	modelObj.SetGroupVersionKind(externalModelGVK)
```
nirrozenbaum: ditto

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -44,32 +45,83 @@ var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
...
+	modelReconciler := &externalModelReconciler{
+		Reader:        k8sClient,
+		modelStore:    modelStore,
+		providerStore: providerStore,
+	}
+	mapProviderToModels := func(ctx context.Context, obj client.Object) []reconcile.Request {
+		providerName := obj.GetName()
+		providerNamespace := obj.GetNamespace()
+		modelList := &unstructured.UnstructuredList{}
```
nirrozenbaum: ditto

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -44,32 +45,83 @@ var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
 
 // ModelProviderResolverFactory defines the factory function for ModelProviderResolverPlugin
 func ModelProviderResolverFactory(name string, _ json.RawMessage, handle framework.Handle) (framework.BBRPlugin, error) {
-	plugin, err := NewModelProviderResolver(handle.ReconcilerBuilder, handle.Client())
+	plugin, err := NewModelProviderResolver(handle)
```
nirrozenbaum: why?
usually we keep handle only in factory functions.
handle is expected to grow (more interfaces) and  we keep the functions with getting the minimal set of args

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -44,32 +45,83 @@ var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
...
+		modelList := &unstructured.UnstructuredList{}
+		modelList.SetGroupVersionKind(externalModelGVK)
+		if err := k8sClient.List(ctx, modelList, client.InNamespace(providerNamespace)); err != nil {
+			log.FromContext(ctx).Error(err, "failed to list ExternalModels for provider mapping",
+				"provider", providerName, "namespace", providerNamespace)
+			return nil
+		}
+		var requests []reconcile.Request
+		for i := range modelList.Items {
+			refs, _, _ := unstructured.NestedSlice(modelList.Items[i].Object, "spec", "externalProviderRefs")
```
nirrozenbaum: I'm waiting for the next cycle for reviewing this part.. need to remove unstructured and use typed clients for our CRDs.
the only place that it's ok to use unstructred for now is for istio resources.

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -96,8 +148,6 @@ func (p *ModelProviderResolverPlugin) WithName(name string) *ModelProviderResolv
 // from the modelInfoStore (populated by ExternalModel reconciler), and writes model, provider
 // and credential reference info to CycleState.
 func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleState *framework.CycleState, request *framework.InferenceRequest) error {
-	logger := log.FromContext(ctx).V(logutil.DEFAULT)
```
nirrozenbaum, line 99: return

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -121,13 +171,12 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 	}
 
 	if !strings.HasSuffix(relativePath, "chat/completions") { // no support for other input types
-		logger.Error(nil, "unsupported route for external model", "model", modelKey.String(), "path", relativePath)
```
nirrozenbaum, line 124: ditto

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -121,13 +171,12 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
...
 		return errcommon.Error{Code: errcommon.BadRequest, Msg: "only /chat/completions input type is supported"}
+
 	}
 
 	// if there's a mismatch it's an error, we don't want to proceed
 	if externalModelInfo.targetModel != model {
-		logger.Error(nil, "model mismatch between request body and ExternalModel", "requestModel", model, "externalModel", externalModelInfo.targetModel)
```
nirrozenbaum, line 130: ditto

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -136,8 +185,10 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 	cycleState.Write(state.ModelKey, externalModelInfo.targetModel)
 	cycleState.Write(state.CredsRefName, externalModelInfo.secretName)
 	cycleState.Write(state.CredsRefNamespace, externalModelInfo.secretNamespace)
+	if len(externalModelInfo.config) > 0 {
+		cycleState.Write(state.ProviderConfigKey, externalModelInfo.config)
+	}
 
-	logger.Info("external model resolved", "model", modelKey.String(), "provider", externalModelInfo.provider)
```
nirrozenbaum, line 140: ditto

pkg/plugins/common/state/state-keys.go
```diff
@@ -22,4 +22,5 @@ const (
 	ModelKey          = "model"
 	CredsRefName      = "credential-ref-name"
 	CredsRefNamespace = "credential-ref-namespace"
+	ProviderConfigKey = "provider-config"
```
nirrozenbaum: what is this key used for?
keep in mind that we're going to add in follow up config also in model. 
and then we'll have:
provider config used as default. 
model config is added to the same map - if it has new keys, map is enriched, if same key like provider, it overrides it.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var externalProviderGVK = schema.GroupVersionKind{
+	Group:   "inference.opendatahub.io",
+	Version: "v1alpha1",
+	Kind:    "ExternalProvider",
+}
```
nirrozenbaum: typed client pls

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+
+type externalProviderReconciler struct {
+	client.Reader
+	store *providerInfoStore
+}
+
+func (r *externalProviderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
```
nirrozenbaum: follow the same logging patterns from model (before you removed lines

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+	secretName, _, _ := unstructured.NestedString(obj.Object, "spec", "auth", "secretRef", "name")
+
+	if providerType == "" || endpoint == "" || secretName == "" {
+		r.store.delete(req.NamespacedName)
+		logger.Info("ExternalProvider missing required fields, removing from store",
+			"provider", providerType, "endpoint", endpoint, "secretName", secretName)
+		return ctrl.Result{}, nil
+	}
+
+	configRaw, _, _ := unstructured.NestedStringMap(obj.Object, "spec", "config")
```
nirrozenbaum: config is optional. where do we check it?

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+	providerType, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
+	endpoint, _, _ := unstructured.NestedString(obj.Object, "spec", "endpoint")
+	secretName, _, _ := unstructured.NestedString(obj.Object, "spec", "auth", "secretRef", "name")
+
+	if providerType == "" || endpoint == "" || secretName == "" {
+		r.store.delete(req.NamespacedName)
+		logger.Info("ExternalProvider missing required fields, removing from store",
+			"provider", providerType, "endpoint", endpoint, "secretName", secretName)
+		return ctrl.Result{}, nil
+	}
```
nirrozenbaum: is this reachable?

pkg/plugins/model-provider-resolver/store.go
```diff
@@ -0,0 +1,61 @@
...
+
+type providerInfo struct {
+	provider        string
+	endpoint        string
+	secretName      string
+	secretNamespace string
+	config          map[string]string
+}
+
+type providerInfoStore struct {
```
nirrozenbaum, line 44: not a big deal, but why not have one store.go file with providers and models with a single lock?
conceptually - providers without models have no sense, and models without providers as well. 
so it might make sense to store them in the same file so a ready can easily understand the relationship (otherwise code is fragmented and harder to follow)

pkg/controller/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-openai
+  namespace: llm
+  labels:
+    app.kubernetes.io/managed-by: ipp-external-provider-reconciler
+    inference.opendatahub.io/external-provider: my-openai
```
nirrozenbaum, line 43: now that CRs example is here, we can easily see that these labels are not needed. owner reference already covers these.
can we remove these labels in the reconciler?

pkg/controller/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```yaml
+apiVersion: networking.istio.io/v1
+kind: ServiceEntry
+metadata:
+  name: my-openai
+  namespace: llm
+  labels:
+    app.kubernetes.io/managed-by: ipp-external-provider-reconciler
+    inference.opendatahub.io/external-provider: my-openai
```
nirrozenbaum, line 70: ditto

pkg/controller/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```yaml
+apiVersion: networking.istio.io/v1
+kind: DestinationRule
+metadata:
+  name: my-openai
+  namespace: llm
+  labels:
+    app.kubernetes.io/managed-by: ipp-external-provider-reconciler
+    inference.opendatahub.io/external-provider: my-openai
```
nirrozenbaum, line 100: ditto

pkg/controller/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: gpt4
+  namespace: llm
+  labels:
+    app.kubernetes.io/managed-by: ipp-external-model-reconciler
+    inference.opendatahub.io/external-model: gpt4
```
nirrozenbaum, line 152: ditto. 

after seeing this in all places - let's get rid of these labels.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Fixed — restored DEFAULT log level, lowercase messages, and deletion log lines.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Fixed — restored DEFAULT log level, lowercase messages, and deletion log lines.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Fixed — restored DEFAULT log level, lowercase messages, and deletion log lines.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
yossiovadia: Fixed — restored DEFAULT log level, lowercase messages, and deletion log lines.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Removed — CRD validation prevents these cases.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Removed — CRD validation prevents these cases.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Removed — CRD validation prevents these cases.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Created #278 to track extending from single to multiple provider refs.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Replaced with `Requeue: true` — lets the controller runtime handle the backoff.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
yossiovadia: Done — renamed to `providerInfo`.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Reverted — `NewModelProviderResolver` now takes `reconcilerBuilder` and `k8sClient` instead of `handle`.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Done — switched to typed clients in latest push.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Done — switched to typed clients in latest push.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Done — switched to typed clients in latest push.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Done — switched to typed clients in latest push.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
yossiovadia: Done — switched to typed clients in latest push.

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: These are the error return paths — they return `ctrl.Result{}` with the error. The code is correct as-is (reconciler returns early on error).

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: These are the error return paths — they return `ctrl.Result{}` with the error. The code is correct as-is (reconciler returns early on error).

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: These are the error return paths — they return `ctrl.Result{}` with the error. The code is correct as-is (reconciler returns early on error).

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: These are the error return paths — they return `ctrl.Result{}` with the error. The code is correct as-is (reconciler returns early on error).

pkg/plugins/common/state/state-keys.go
yossiovadia: Used by downstream plugins (e.g., api-translation for Vertex project/location). When model-level config is added (#260), the merge logic (model overrides provider) will happen here before writing to CycleState.

pkg/plugins/model-provider-resolver/plugin.go
nirrozenbaum: @yossiovadia can we switch to typed client in this PR? 
I don’t see a point to review and merge controllers code that immediately changes

pkg/plugins/common/state/state-keys.go
nirrozenbaum: right. so can we use model-config instead? 
this config is not planned to be provider config but model-config (which uses default values from provider)

pkg/controller/README.md
yossiovadia: These labels are in the controller reconcilers (`pkg/controller/externalprovider/reconciler.go` and `externalmodel/reconciler.go`) from PR #183 (already merged). The `managedByPredicate` in `SetupWithManager` uses them to filter owned resources.

Removing labels means also updating the predicate in the merged controller code. Should I do that as part of this PR, or as a separate follow-up on main?

pkg/plugins/model-provider-resolver/plugin.go
yossiovadia: Done in the latest push — store reconcilers now use typed `inferencev1alpha1.ExternalProvider` and `ExternalModel`. Legacy `maas.opendatahub.io` reconciler stays unstructured (types are in a different repo).

pkg/plugins/common/state/state-keys.go
yossiovadia: Makes sense — will rename to `ModelConfigKey` since the merged config (provider defaults + model overrides) is model-scoped.

pkg/plugins/model-provider-resolver/store.go
yossiovadia: Fair point. Will merge into a single `store.go` with both provider and model maps under one lock.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
yossiovadia: Config is optional in the CRD (`+optional`). The reconciler reads whatever is there — if empty, `provider.Spec.Config` is `nil`, and `providerInfo.config` stays `nil`. Downstream, ProcessRequest only writes to CycleState if `len(config) > 0`. No explicit check needed.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
yossiovadia: Which line are you referring to? The typed reconciler reads from `provider.Spec` fields directly — all required fields are validated by the CRD.

pkg/controller/README.md
yossiovadia: See reply on the first labels comment — these are in merged #183 code.

pkg/controller/README.md
yossiovadia: See reply on the first labels comment — these are in merged #183 code.

pkg/controller/README.md
yossiovadia: See reply on the first labels comment — these are in merged #183 code.

yossiovadia:
Replaced by smaller incremental PRs starting with #285. See plan shared with Nir.

PR #273, opened 2026-05-14
fix(nemo): avoid per-request message copy in extractOpenAIMessages and return the original

liavweiss:
## Summary

- Return the original `[]any` slice directly from `extractOpenAIMessages` instead of copying each message into a new `map[string]string` on every request
- On long multi-turn conversations, the previous approach allocated a new slice and iterated over every message to extract only `role` and `content` - this is expensive when the chat history grows
- NeMo ignores extra fields (`tool_calls`, `function_call`, `name`), so forwarding the original message objects is safe and produces identical behavior
- Return types updated from `[]map[string]string` to `[]any` for `extractMessages`, `extractOpenAIMessages`, and `extractMCPArguments`
- `extractMCPArguments` still constructs messages (MCP is a different payload format) but returns `[]any` for type consistency

## Tested

- All unit tests pass (`go vet` + `go test ./pkg/plugins/nemo/...`)
- E2E tested on Kind cluster: Envoy -> BBR (ext-proc) -> NeMo guardrails
  - Safe message -> 200 (allowed)
  - Blocked message -> 403 (blocked)
  - No messages / empty messages -> 200 (no-op, NeMo not called)
  - Messages with extra fields (`tool_calls`, `name`) -> 200 (forwarded as-is, NeMo ignores extras)

## Related
#272

## Summary by CodeRabbit

* **Refactor**
  * Internal improvements to message handling and request processing in the plugin system. No visible changes to end-user functionality.

### DIFF
--- a/pkg/plugins/nemo/request_guard.go
+++ b/pkg/plugins/nemo/request_guard.go
@@ -138,7 +138,7 @@ func (p *NemoRequestGuardPlugin) ProcessRequest(ctx context.Context, _ *framewor
 //     all string argument values into a single user message.
 //
 // Returns (nil, nil) when no content is found.
-func extractMessages(body map[string]any) ([]map[string]string, error) {
+func extractMessages(body map[string]any) ([]any, error) {
 	if raw, ok := body["messages"]; ok {
 		return extractOpenAIMessages(raw)
 	}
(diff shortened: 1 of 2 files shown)


nirrozenbaum (approved):
nice!

I assume it was tested with Trusty?

/lgtm
/approve

liavweiss:
> nice!
> 
> I assume it was tested with Trusty?
> 
> /lgtm /approve

Yes of course.

PR #285, opened 2026-05-18
feat: add ExternalProvider store and typed reconciler

yossiovadia:
## Summary

Adds a provider info store and typed reconciler for `inference.opendatahub.io` ExternalProvider CRDs. This is the first of a series of small incremental PRs to wire ExternalProvider/ExternalModel into the model-provider-resolver plugin.

**4 files, 124 lines production code, 193 lines tests.**

## What this does

- `provider_store.go` — thread-safe in-memory store for provider info (endpoint, credentials, config)
- `external_provider_reconciler.go` — typed client reconciler that watches ExternalProvider CRDs and populates the store
- Tests for both

## What this does NOT do (follow-up PRs)

- Does not register the reconciler in the plugin factory (next PR)
- Does not modify the model reconciler to use the provider store (next PR)
- No RBAC, no Helm changes, no E2E infra changes

## Test plan

- [x] Unit tests pass
- [x] Lint clean (0 issues)

## Risk analysis

- **Risk rating**: 1
- **Why**: Adds new files only, no existing code modified. The store and reconciler are not wired into anything yet — zero runtime impact until the next PR registers them.

cc @nirrozenbaum

## Summary by CodeRabbit

* **New Features**
  * External provider resources can now be automatically reconciled and managed within the system.
  * Provider configurations are persisted in an internal store for dynamic access.

* **Tests**
  * Added comprehensive unit tests for provider reconciliation scenarios including creation, updates, and deletion.
  * Added tests validating provider configuration storage and retrieval.

PR #286, opened 2026-05-18
feat: typed ExternalModel reconciler with cross-watch and config/auth override

yossiovadia:
## Summary

Depends on #285. Updates the model-provider-resolver plugin to use typed clients for `inference.opendatahub.io` ExternalModels with provider cross-watch, plus legacy backward compatibility for `maas.opendatahub.io`.

**4 files changed, 121 insertions, 46 deletions.**

## What this does

- Typed ExternalModel reconciler cross-references provider store for credential/config resolution
- Cross-watch: ExternalProvider changes trigger re-reconciliation of dependent models
- Legacy `maas.opendatahub.io` ExternalModel reconciler (unstructured, backward compat)
- `ModelConfigKey` for config propagation via CycleState
- Scheme registration in factory for typed client support
- Registers all three reconcilers (provider, model, legacy) in the plugin

## Test plan

- [x] Unit tests pass, lint clean (0 issues)

## Risk analysis

- **Risk rating**: 2
- **Why**: Modifies existing plugin.go (factory + reconciler registration). The legacy reconciler ensures backward compatibility with existing E2E tests that use maas.opendatahub.io CRDs.

Depends on: #285
cc @nirrozenbaum

## Summary by CodeRabbit

* **New Features**
  * Store per-model configuration merged from providers and per-model overrides.

* **Improvements**
  * Reconciliation now uses typed provider/model APIs for more reliable provider discovery and credential handling.
  * Request validation tightened: only chat/completions paths accepted and incoming request model must match the target model.
  * Added a state key to persist resolved model configuration.

* **Refactor**
  * Resolver initialization rewritten to use a shared in-memory store and propagate provider changes to affected models.

### DIFF
--- a/pkg/plugins/model-provider-resolver/external_model_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_model_reconciler.go
@@ -80,10 +80,7 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1al
 		return nil, false
 	}
 
-	config := providerInfo.config
-	if len(ref.Config) > 0 {
-		config = mergeConfig(providerInfo.config, ref.Config)
-	}
+	config := mergeConfig(providerInfo.config, ref.Config)
 
 	secretName := providerInfo.secretName
 	secretNamespace := providerInfo.secretNamespace
--- a/pkg/plugins/model-provider-resolver/external_model_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_model_reconciler.go
@@ -19,64 +19,97 @@ package model_provider_resolver
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/observability/logging"
+
+	inferencev1alpha1 "github.com/opendatahub-io/ai-gateway-payload-processing/api/inference/v1alpha1"
 )
 
-// externalModelGVK is the GroupVersionKind for ExternalModel CRD.
-var externalModelGVK = schema.GroupVersionKind{
-	Group:   "maas.opendatahub.io",
-	Version: "v1alpha1",
-	Kind:    "ExternalModel",
-}
+const providerRequeueDelay = 5 * time.Second
 
-// externalModelReconciler watches ExternalModel CRDs (via unstructured client)
-// and updates the model store with provider and credential information.
+// externalModelReconciler watches inference.opendatahub.io ExternalModel CRDs
+// and resolves provider info from the provider store.
 type externalModelReconciler struct {
 	client.Reader
 	store *infoStore
 }
 
-// Reconcile handles create/update/delete events for ExternalModel resources.
-// The ExternalModel CR name is used as the model key in the store, matching
-// the model name in inference request bodies.
 func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx).V(logutil.DEFAULT)
 	logger.Info("reconciling ExternalModel", "name", req.Name, "namespace", req.Namespace)
 
-	obj := &unstructured.Unstructured{}
-	obj.SetGroupVersionKind(externalModelGVK)
-
-	err := r.Get(ctx, req.NamespacedName, obj)
+	model := &inferencev1alpha1.ExternalModel{}
+	err := r.Get(ctx, req.NamespacedName, model)
 	if err != nil && !errors.IsNotFound(err) {
 		return ctrl.Result{}, fmt.Errorf("unable to get ExternalModel: %w", err)
 	}
 
-	if errors.IsNotFound(err) || !obj.GetDeletionTimestamp().IsZero() {
+	if errors.IsNotFound(err) || !model.GetDeletionTimestamp().IsZero() {
 		r.store.deleteModel(req.NamespacedName)
 		logger.Info("ExternalModel removed from store", "name", req.Name, "namespace", req.Namespace)
 		return ctrl.Result{}, nil
 	}
 
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
-
-	// targetModel is the model that will be used in the request body when getting inference requests.
-	info := &externalModelInfo{
-		provider:        provider,
-		targetModel:     targetModel,
-		secretName:      credsName,
-		secretNamespace: req.Namespace, // secret namespace is always the namespace of the ExternalModel
+	// Use the first ref whose provider is available in the store.
+	for _, ref := range model.Spec.ExternalProviderRefs {
+		info, found := r.resolveRef(req.Namespace, ref)
+		if !found {
+			continue
+		}
+		r.store.addOrUpdateModel(req.NamespacedName, info)
+		logger.Info("updated model store", "provider", info.provider, "targetModel", info.targetModel)
+		return ctrl.Result{}, nil
 	}
-	r.store.addOrUpdateModel(req.NamespacedName, info)
 
-	logger.Info("updated model store", "provider", provider, "targetModel", targetModel)
-	return ctrl.Result{}, nil
+	logger.Info("no ExternalProvider available for any ref, requeuing")
+	return ctrl.Result{RequeueAfter: providerRequeueDelay}, nil
 }
+
+// resolveRef resolves a single ExternalProviderRef to model info.
+// Returns (nil, false) if the provider is not yet available in the store.
+func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1alpha1.ExternalProviderRef) (*externalModelInfo, bool) {
+	providerKey := types.NamespacedName{Namespace: namespace, Name: ref.Ref.Name}
+	providerInfo, found := r.store.getProvider(providerKey)
+	if !found {
+		return nil, false
+	}
+
+	config := providerInfo.config
+	if len(ref.Config) > 0 {
+		config = mergeConfig(providerInfo.config, ref.Config)
+	}
+
+	secretName := providerInfo.secretName
+	secretNamespace := providerInfo.secretNamespace
+	if ref.Auth != nil {
+		secretName = ref.Auth.SecretRef.Name
+		secretNamespace = namespace
+	}
+
+	return &externalModelInfo{
+		provider:        providerInfo.provider,
+		targetModel:     ref.TargetModel,
+		secretName:      secretName,
+		secretNamespace: secretNamespace,
+		config:          config,
+	}, true
+}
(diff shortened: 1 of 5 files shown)


nirrozenbaum:
@yossiovadia I think this PR has to be changed. 
the legacy group reconciler should just create the new CRDs (and that will trigger the new logic).
your thoughts?

yossiovadia:
Done — removed the legacy  store reconciler. The legacy-to-new-CR conversion is now entirely in #292 (migration controller).

What changed:
- Removed `legacyExternalModelReconciler`, `legacyExternalModelGVK`, and the legacy registration block from `plugin.go`
- The PR is now purely: typed ExternalModel reconciler + cross-watch + config/auth override
- All tests pass, lint clean

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -22,28 +22,32 @@ import (
...
 	ModelProviderResolverPluginType = "model-provider-resolver"
 )
 
-// compile-time type validation
 var _ framework.RequestProcessor = &ModelProviderResolverPlugin{}
 
-// ModelProviderResolverFactory defines the factory function for ModelProviderResolverPlugin
+// ModelProviderResolverFactory defines the factory function for ModelProviderResolverPlugin.
 func ModelProviderResolverFactory(name string, _ json.RawMessage, handle framework.Handle) (framework.BBRPlugin, error) {
+	utilruntime.Must(inferencev1alpha1.AddToScheme(handle.Client().Scheme()))
```
nirrozenbaum: nit: better to do that in NewModelProviderResolver, isn't it?
if in some tests we initialize the plugin through code.

not a blocker

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -52,55 +56,72 @@ func ModelProviderResolverFactory(name string, _ json.RawMessage, handle framewo
...
-}
+func (p *ModelProviderResolverPlugin) TypedName() plugin.TypedName { return p.typedName }
 
-// WithName sets the name of the plugin instance.
 func (p *ModelProviderResolverPlugin) WithName(name string) *ModelProviderResolverPlugin {
 	p.typedName.Name = name
 	return p
 }
 
-// ProcessRequest reads the model name from the request body, resolves the provider
```
nirrozenbaum, line 95: in this file the current PR removes all doc commnets which were kinda useful for understanding the code. 
not a blocker, but would be nice to put the comments back :)

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -116,37 +137,33 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
...
 
-	if !strings.HasSuffix(relativePath, "chat/completions") { // no support for other input types
-		logger.Error(nil, "unsupported route for external model", "model", modelKey.String(), "path", relativePath)
+	if !strings.HasSuffix(relativePath, "chat/completions") {
 		return errcommon.Error{Code: errcommon.BadRequest, Msg: "only /chat/completions input type is supported"}
 	}
 
-	// if there's a mismatch it's an error, we don't want to proceed
 	if externalModelInfo.targetModel != model {
-		logger.Error(nil, "model mismatch between request body and ExternalModel", "requestModel", model, "externalModel", externalModelInfo.targetModel)
```
nirrozenbaum, line 130: need to put this back, otherwise we won't see any error in the logs

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,87 @@ package model_provider_resolver
...
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
+	// CRD validation ensures at least one ref (MinItems=1)
+	ref := model.Spec.ExternalProviderRefs[0]
+
+	providerKey := types.NamespacedName{Namespace: req.Namespace, Name: ref.Ref.Name}
+	providerInfo, found := r.store.getProvider(providerKey)
+	if !found {
+		logger.Info("ExternalProvider not yet available, requeuing", "provider", ref.Ref.Name)
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
nirrozenbaum: pls remove the magic number 5 sec

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,87 @@ package model_provider_resolver
...
+	merged := make(map[string]string, len(providerConfig)+len(modelConfig))
+	for k, v := range providerConfig {
+		merged[k] = v
+	}
+	for k, v := range modelConfig {
+		merged[k] = v
+	}
+	return merged
+}
+
```
nirrozenbaum: we can remove this function.
no need to create a new map with len(p-config + m-config).. len also could be wrong (could be double if all keys are the same).

we can leave only the model config overriding the provider config:

```suggestion
for k, v := range modelConfig {
	config[k] = v
}
```

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,87 @@ package model_provider_resolver
...
 		r.store.deleteModel(req.NamespacedName)
 		logger.Info("ExternalModel removed from store", "name", req.Name, "namespace", req.Namespace)
 		return ctrl.Result{}, nil
 	}
 
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
+	// CRD validation ensures at least one ref (MinItems=1)
+	ref := model.Spec.ExternalProviderRefs[0]
```
nirrozenbaum: can we generalize this? 
instead of doing the function only for refs[0], extract the code below to a "handle" function and iterate over all refs, call "handle" for each of them.
it should be a 5 lines changes.. no point in deferring to a separate PR

nirrozenbaum (commented):
@yossiovadia left few minor comments, nothing serious.
the only important one is the generalize the refs handling (multiple refs and not only the first one).
additionally need to see why e2e tests fail (cannot be merged until fixed)

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,97 @@ package model_provider_resolver
...
-	info := &externalModelInfo{
-		provider:        provider,
-		targetModel:     targetModel,
-		secretName:      credsName,
-		secretNamespace: req.Namespace, // secret namespace is always the namespace of the ExternalModel
+	secretName := providerInfo.secretName
+	secretNamespace := providerInfo.secretNamespace
+	if ref.Auth != nil {
+		secretName = ref.Auth.SecretRef.Name
+		secretNamespace = namespace
```
nirrozenbaum, line 89: nit: provider/model are in the same ns, this is redundant.

```suggestion
```

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -116,37 +144,36 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
...
 	}
 
-	// info of external model written to cycle state for next plugins
 	cycleState.Write(state.ProviderKey, externalModelInfo.provider)
 	cycleState.Write(state.ModelKey, externalModelInfo.targetModel)
 	cycleState.Write(state.CredsRefName, externalModelInfo.secretName)
 	cycleState.Write(state.CredsRefNamespace, externalModelInfo.secretNamespace)
+	if len(externalModelInfo.config) > 0 {
+		cycleState.Write(state.ModelConfigKey, externalModelInfo.config)
+	}
```
nirrozenbaum, line 169: it would be easier to maintain the case where the map is always written to avoid the conditionals in multiple plugins. we can have empty map in no config was defined.

nirrozenbaum: that probably means need to update that in the addOrUpdateProvider function (check if config is nil then create an empty map in the store.go)

test/e2e/e2e_test.go
```diff
@@ -299,6 +316,11 @@ var _ = ginkgo.Describe("IPP Plugin Chain", ginkgo.Label("e2e"), func() {
 		for _, p := range providers {
 			p := p
 
+			// Simulator's anthropic provider has auto-tool disabled (Claude Code compat)
+			if p.Provider == "anthropic" {
+				continue
+			}
```
nirrozenbaum: should this be fixed in the simulator?

nirrozenbaum (approved):
/lgtm

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,97 @@ package model_provider_resolver
...
+	providerKey := types.NamespacedName{Namespace: namespace, Name: ref.Ref.Name}
+	providerInfo, found := r.store.getProvider(providerKey)
+	if !found {
+		return nil, false
+	}
+
+	config := providerInfo.config
+	if len(ref.Config) > 0 {
+		config = mergeConfig(providerInfo.config, ref.Config)
+	}
```
nirrozenbaum: when ref.Config is empty, config aliases providerInfo.config directly. Any downstream mutation of the config map via ModelConfigKey would corrupt the shared provider map and leak across models.

Fix: call mergeConfig unconditionally instead of the if len(ref.Config) > 0 branch.

nirrozenbaum (commented):
@yossiovadia thanks, overall looks good. 
few notes:
there's an issue in config when it's empty, worth fixing before merge otherwise this bug might bite us later.

few other points:
- I noticed apiFormat is not handled at all. not saved in datastore nor in CycleState. can we handle that to prepare the group for selecting the right api translator based on the apiFormat?
- The stripped doc comments in plugin.go were useful for understanding the code flow (e.g., why early returns are not errors, what gets written to CycleState). Would be nice to put them back.
- Config should always be written to CycleState (even as an empty map) to avoid downstream plugins needing nil checks. This means addOrUpdateProvider should initialize an empty map when spec.config is nil.

yossiovadia:
@nirrozenbaum Pushed fixes for #1 (config alias bug — mergeConfig called unconditionally now) and #3 (doc comments restored).

For #2 (apiFormat in store + CycleState) and #4 (always write config to CycleState) — I'd like to do those as a quick follow-up PR after merge to keep this one focused. Sound good?

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -19,64 +19,94 @@ package model_provider_resolver
...
-	provider, _, _ := unstructured.NestedString(obj.Object, "spec", "provider")
-	targetModel, _, _ := unstructured.NestedString(obj.Object, "spec", "targetModel")
-	credsName, _, _ := unstructured.NestedString(obj.Object, "spec", "credentialRef", "name")
+	// Use the first ref whose provider is available in the store.
+	for _, ref := range model.Spec.ExternalProviderRefs {
+		info, found := r.resolveRef(req.Namespace, ref)
+		if !found {
+			continue
+		}
+		r.store.addOrUpdateModel(req.NamespacedName, info)
```
nirrozenbaum, line 65: this is a bug. if there are multiple providers the first one in the slice will be in the info. 
we should have in the store per provider info. 

can be done in a follow up together with the other leftovers

nirrozenbaum (approved):
/lgtm
/approve

let's handle in a quick follow up the PR leftovers:

- apiFormat in store + CycleState (only store in CycleState, can be consumed in api translator in next PRs).
- always write config to CycleState. if it doesn't exist create empty map.
- add weight to provider ref. if no weight exist default to weight = 1. should be stored in store and used by plugin resolver to select the provider based on the weight.
- update the code in the reconciler to support multiple provider refs, adding all entries to the store (including weight)/

PR #291, opened 2026-05-19
feat: install inference.opendatahub.io CRDs in E2E infra + read RBAC

yossiovadia:
## Summary

Installs `inference.opendatahub.io` CRDs and adds read-only RBAC so the store reconcilers (from #285, #286) can watch ExternalProvider and ExternalModel resources.

**4 files, 22 lines.**

## What this does

- `setup-kind.sh` — install inference CRDs from `config/crd/bases/`
- `entrypoint.sh` — install inference CRDs in E2E test container
- `Dockerfile.e2e` — embed CRD files in test image
- `rbac.yaml` — add `inference.opendatahub.io` get/list/watch for externalproviders and externalmodels

## Why needed

Without this, the store reconcilers crash in CI:
```
"if kind is a CRD, it should be installed before calling Start"
"no matches for kind ExternalProvider in version inference.opendatahub.io/v1alpha1"
```

## Risk analysis

- **Risk rating**: 1
- **Why**: Only adds CRD installation and read-only RBAC. No code changes.

cc @nirrozenbaum

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -43,6 +43,7 @@ WORKDIR /e2e
 COPY --from=builder /e2e-tests.test /e2e/e2e-tests.test
 COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
 COPY test/e2e/scripts/entrypoint.sh /e2e/entrypoint.sh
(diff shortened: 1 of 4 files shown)


nirrozenbaum (approved):
/lgtm

PR #293, opened 2026-05-19
fix: use real Istio and Gateway API CRDs in envtest

yossiovadia:
## Summary

- Replace minimal CRD stubs (`x-kubernetes-preserve-unknown-fields: true`) with real upstream Istio and Gateway API CRDs
- Add `hack/download-test-crds.sh` to download and cache CRDs from upstream releases (Istio 1.25.2, Gateway API v1.3.0)
- Wire `download-test-crds` as a `make test-unit` prerequisite so CRDs are always available before envtest runs
- Delete `testdata/istio-crds/` and `testdata/gateway-api-crds/` stub directories from controller packages

This ensures envtest validates resources against real CRD schemas, matching production behavior more closely.

Closes #239

## Test plan

- [x] `make test-unit` — all tests pass with real CRDs (externalprovider: 87.9%, externalmodel: 84.5%)
- [x] `make lint` — 0 issues
- [x] Downloaded CRDs are gitignored and cached (re-run is a no-op)


nirrozenbaum (approved):
/lgtm
/approve

PR #294, opened 2026-05-19
fix: use oc CLI instead of kubectl in E2E container

yossiovadia:
## Summary

- Replace `kubectl` with `oc` in the E2E Dockerfile and entrypoint script for OpenShift/Konflux compatibility
- Download `oc` from Red Hat mirror (`stable-4.17` channel) — the tarball includes both `oc` and `kubectl`
- Go test helpers use `KUBECTL_BIN` env var (default: `oc`) so local runs can override with `kubectl`
- `setup-kind.sh` is unchanged — it targets Kind clusters where `kubectl` is correct

Per DevTestOps team request to use `oc` in E2E test containers.

## Test plan

- [x] `go vet ./test/e2e/` — clean
- [x] `make lint` — 0 issues
- [ ] Verify Dockerfile builds with `oc` download (CI)
- [ ] E2E on Konflux with `oc` (CI)

## Summary by CodeRabbit

* **New Features**
  * Added support for overriding the kubectl binary via `KUBECTL_BIN` environment variable in E2E tests.
  * E2E container runtime now includes the OpenShift CLI (`oc`) binary.

* **Bug Fixes**
  * Improved error reporting and debugging output for kubectl command failures in E2E tests.
  * E2E setup now uses OpenShift CLI for cluster connectivity verification and CRD management.

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -11,7 +11,7 @@
 
 ARG GOLANG_VERSION=1.25
 
-## Stage 1: Build the test binary and download kubectl
+## Stage 1: Build the test binary and download oc
 FROM registry.access.redhat.com/ubi9/go-toolset:$GOLANG_VERSION AS builder
 ARG OCP_VERSION=4.21.15
 ARG CGO_ENABLED=1
@@ -33,11 +33,12 @@ RUN CGO_ENABLED=${CGO_ENABLED} GOEXPERIMENT=strictfipsruntime GOOS=${TARGETOS} G
     go test -c -o /e2e-tests.test ./test/e2e/ \
     -ldflags="-s -w"
 
-# Download FIPS-compliant kubectl from the OpenShift client tarball (rhel9, CGO+OpenSSL).
+# Download FIPS-compliant oc and kubectl from the OpenShift client tarball.
 RUN curl -sL "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_VERSION}/openshift-client-linux-${TARGETARCH}-rhel9-${OCP_VERSION}.tar.gz" | \
     tar -xz -C /tmp/ oc kubectl && \
+    mv /tmp/oc /usr/local/bin/oc && \
     mv /tmp/kubectl /usr/local/bin/kubectl && \
-    chmod +x /usr/local/bin/kubectl
+    chmod +x /usr/local/bin/oc /usr/local/bin/kubectl
 
 ## Stage 2: Runtime
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8@sha256:ae09ecc3d754bc1726cbda3e2599cc7839e09fe1cc547ce173cf669b645be3cc
@@ -45,6 +46,7 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8@sha256:ae09ecc3d754bc1726cb
 WORKDIR /e2e
 
 COPY --from=builder /e2e-tests.test /e2e/e2e-tests.test
+COPY --from=builder /usr/local/bin/oc /usr/local/bin/oc
 COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
 COPY test/e2e/scripts/entrypoint.sh /e2e/entrypoint.sh
 COPY config/crd/bases/ /e2e/crds/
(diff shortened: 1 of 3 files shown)


noyitz:
This conflicts with PR #275 which also replaces kubectl in the Dockerfile. I approved #275 first since it includes CGO + FIPS alignment. 

The Go-side changes here are good though — the configurable `KUBECTL_BIN` env var and `oc` in entrypoint.sh are useful additions. Can you rebase on top of #275 once it merges and keep just the Go/entrypoint changes?

yossiovadia:
@nirrozenbaum  plz review this small PR

Dockerfile.e2e
```diff
@@ -33,18 +33,20 @@ RUN CGO_ENABLED=${CGO_ENABLED} GOEXPERIMENT=strictfipsruntime GOOS=${TARGETOS} G
...
-    chmod +x /usr/local/bin/kubectl
+    chmod +x /usr/local/bin/oc /usr/local/bin/kubectl
 
 ## Stage 2: Runtime
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8@sha256:ae09ecc3d754bc1726cbda3e2599cc7839e09fe1cc547ce173cf669b645be3cc
 
 WORKDIR /e2e
 
 COPY --from=builder /e2e-tests.test /e2e/e2e-tests.test
+COPY --from=builder /usr/local/bin/oc /usr/local/bin/oc
```
noyitz, line 49: Do we need both `oc` and `kubectl` in the image? `oc` is a superset of `kubectl` — having both adds ~140MB. If we're switching to `oc`, we could just symlink `kubectl -> oc` instead of copying both binaries.

test/e2e/e2e_suite_test.go
```diff
@@ -34,6 +34,7 @@ var (
 	gatewaySvcName string
 	simulatorEP    string
 	curlTimeout    = 30 * time.Second
+	kubectlBin     = envOr("KUBECTL_BIN", "kubectl")
```
noyitz, line 37: This defaults to `kubectl` but `entrypoint.sh` was changed to hardcode `oc`. They should be consistent — either default to `oc` here too, or keep `kubectl` in the entrypoint.

test/e2e/scripts/entrypoint.sh
```diff
@@ -33,16 +33,16 @@ if [[ ! -f "$KUBECONFIG" ]]; then
 fi
 
 echo "Verifying cluster connectivity..."
-if ! kubectl cluster-info --request-timeout=10s >/dev/null 2>&1; then
+if ! oc cluster-info --request-timeout=10s >/dev/null 2>&1; then
```
noyitz, line 36: On Kind (GitHub Actions CI), `oc` is not installed — only `kubectl`. Since the entrypoint now hardcodes `oc`, the Kind CI will fail unless `oc` is in the image (it is now via the Dockerfile change) or the CI sets `KUBECTL_BIN=kubectl`. Worth adding a note or making sure `ci-e2e.yaml` sets this env var.

Dockerfile.e2e
yossiovadia: Fair point on the size. The oc tarball includes both anyway so we're not downloading extra — just extracting both. A symlink would save the copy but oc and kubectl can behave differently for some operations. Not blocking, can optimize later if image size becomes a concern.

test/e2e/e2e_suite_test.go
yossiovadia: Intentional — two different execution contexts:
- `entrypoint.sh` runs inside the Docker container (Jenkins/Konflux `podman run`) where `oc` is installed
- Go tests run directly on the GitHub Actions runner where only `kubectl` is available

Both paths work — Konflux E2E (container) and GH Actions E2E (runner) both pass. I'll add a comment to make this clearer.

test/e2e/scripts/entrypoint.sh
yossiovadia: This isn't an issue in practice — the entrypoint only runs inside the container (via `podman run`), not on the CI runner. The GH Actions CI runs `go test` directly which uses `kubectl` via the `KUBECTL_BIN` default. Both E2E paths pass (Konflux container + GH Actions runner).

nirrozenbaum:
@yossiovadia rebase please.
@noyitz what's the status of this PR? what's missing to merge?

yossiovadia:
Superseded — rebased cleanly as new PR from current main.

PR #299, opened 2026-06-01
feat: legacy migration controller — convert maas.opendatahub.io to inference.opendatahub.io CRs

yossiovadia:
## Summary

New controller in `pkg/controller/legacymigration/` that watches legacy `maas.opendatahub.io/v1alpha1` ExternalModel CRs and creates corresponding `inference.opendatahub.io/v1alpha1` ExternalProvider + ExternalModel CRs.

**Flow:**
```
Old CR (maas.opendatahub.io) → migration controller creates new CRs
  → existing controllers create HTTPRoute + Service + Istio resources
```

**Design choice:** one ExternalProvider per old CR (no deduplication). See [#292 comment](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/292#issuecomment-4593548433) for discussion on shared provider alternative — happy to switch if preferred.

**Safety:** migration controller only updates CRs it created (checks `managed-by: ipp-legacy-migration` label). Manually created CRs are never overwritten.

## Size note

~130 lines of production code. The rest is envtest test suite (6 test cases, 79% coverage) + license headers. Self-contained new package with no changes to existing code.

Closes #292

## Test plan

- [x] `make test-unit` — all tests pass (legacymigration: 79% coverage)
- [x] `make lint` — 0 issues
- [x] TestReconcile_CreatesNewCRs — old CR creates new ExternalProvider + ExternalModel
- [x] TestReconcile_AnthropicProvider — apiFormat mapping (anthropic → anthropic)
- [x] TestReconcile_UpdateLegacyCR — endpoint change propagates to new provider
- [x] TestReconcile_DeleteLegacyCR — new CRs persist after old CR deletion
- [x] TestReconcile_DoesNotOverwriteManualCRs — manually created CRs are safe
- [x] TestMapProviderToAPIFormat — table-driven test for all provider mappings

## Summary by CodeRabbit

* **New Features**
  * Added automatic migration of legacy resources to the new inference API format.

* **Tests**
  * Added comprehensive test coverage and infrastructure for validating legacy migration functionality.

### DIFF
--- a/pkg/controller/legacymigration/reconciler.go
+++ b/pkg/controller/legacymigration/reconciler.go
@@ -1,0 +1,208 @@
+/*
+Copyright 2026 The opendatahub.io Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package legacymigration
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	inferencev1alpha1 "github.com/opendatahub-io/ai-gateway-payload-processing/api/inference/v1alpha1"
+)
+
+const (
+	labelManagedBy    = "app.kubernetes.io/managed-by"
+	labelMigratedFrom = "inference.opendatahub.io/migrated-from"
+	managedByValue    = "ipp-legacy-migration"
+)
+
+var LegacyExternalModelGVK = schema.GroupVersionKind{
+	Group:   "maas.opendatahub.io",
+	Version: "v1alpha1",
+	Kind:    "ExternalModel",
+}
+
+//+kubebuilder:rbac:groups=maas.opendatahub.io,resources=externalmodels,verbs=get;list;watch
+//+kubebuilder:rbac:groups=inference.opendatahub.io,resources=externalproviders,verbs=get;list;watch;create;update
+//+kubebuilder:rbac:groups=inference.opendatahub.io,resources=externalmodels,verbs=get;list;watch;create;update
+
+// Reconciler watches legacy maas.opendatahub.io ExternalModel CRs and creates
+// corresponding inference.opendatahub.io ExternalProvider + ExternalModel CRs.
+// The new CRs trigger the existing controller flow (HTTPRoute, Service, etc.).
+type Reconciler struct {
+	client.Client
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("reconciling legacy ExternalModel", "name", req.Name, "namespace", req.Namespace)
+
+	old := &unstructured.Unstructured{}
+	old.SetGroupVersionKind(LegacyExternalModelGVK)
+	if err := r.Get(ctx, req.NamespacedName, old); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if !old.GetDeletionTimestamp().IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	providerName, _, _ := unstructured.NestedString(old.Object, "spec", "provider")
+	endpoint, _, _ := unstructured.NestedString(old.Object, "spec", "endpoint")
+	targetModel, _, _ := unstructured.NestedString(old.Object, "spec", "targetModel")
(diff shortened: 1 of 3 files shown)


pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+		return "anthropic"
+	case "azure", "azure-openai":
+		return "azure-openai"
+	case "vertex-openai":
+		return "vertex-openai"
+	case "bedrock", "aws-bedrock":
+		return "bedrock-openai"
+	default:
+		return "openai"
+	}
```
nirrozenbaum: we should document/align on the naming convenitions.
is it openai? is it chat-completions? is it openai-chat?

pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if err := r.applyProvider(ctx, desiredProvider); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to apply ExternalProvider: %w", err)
+	}
+
+	if err := r.applyModel(ctx, desiredModel); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to apply ExternalModel: %w", err)
+	}
+
```
nirrozenbaum, line 136: should we add OwnerReference?

pkg/controller/legacymigration/reconciler.go
yossiovadia: Good call. Currently `pkg/plugins/common/provider/provider.go` defines: `openai`, `anthropic`, `azure-openai`, `vertex-openai`, `bedrock-openai`. These are provider identifiers though, not API format names.

The `apiFormat` field on ExternalProviderRef is described in the CRD as: *"how requests/responses are translated"* — so it maps to the api-translation plugin's translator names.

I'll align with whatever we settle on. Should we use the same values as provider.go (`openai`, `anthropic`, etc.) or introduce separate format names? Happy to create an issue to track the naming convention if you want to decide separately and not block this PR.

pkg/controller/legacymigration/reconciler.go
yossiovadia: Yes, I think we should. Adding OwnerReference from old CR to new CRs means:
- Delete old CR → new CRs auto-cleaned up (cascade)
- Clean state, no orphans

The alternative (no ownership, current behavior) leaves new CRs around after old CR deletion — user would need to clean up manually.

I'll add it. One detail: since the old CR is unstructured, I'll set the OwnerReference manually (like we do in `externalprovider/reconciler.go` with `setUnstructuredOwner`).

pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,217 @@
...
+	if apierrors.IsNotFound(err) {
+		log.FromContext(ctx).Info("creating ExternalProvider", "name", desired.Name)
+		return r.Create(ctx, desired)
+	}
+	if err != nil {
+		return err
+	}
+	if !isManagedByMigration(existing.Labels) {
+		return nil
+	}
```
nirrozenbaum, line 155: this one worth a log line. we could end up creating a model from old gvk that points to "new provider" and this would return without creating it. 
the same could happen for models. 
at least we should have visibility on that.

pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,217 @@
...
+	if apierrors.IsNotFound(err) {
+		log.FromContext(ctx).Info("creating ExternalModel", "name", desired.Name)
+		return r.Create(ctx, desired)
+	}
+	if err != nil {
+		return err
+	}
+	if !isManagedByMigration(existing.Labels) {
+		return nil
+	}
```
nirrozenbaum, line 177: ditto

pkg/controller/legacymigration/reconciler.go
nirrozenbaum: I think we should create new names, cause provider names cannot identify api format (e.g., openai maps to chat completions or to responses?). 
let's go with:
openai-chat
messages
bedrock-messages (future, required cause bedrock messages is different than pure messages).

in this code we always map to openai-chat unless the provider was anthropic than we map to messages.

pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+var LegacyExternalModelGVK = schema.GroupVersionKind{
+	Group:   "maas.opendatahub.io",
+	Version: "v1alpha1",
+	Kind:    "ExternalModel",
+}
+
+//+kubebuilder:rbac:groups=maas.opendatahub.io,resources=externalmodels,verbs=get;list;watch
+//+kubebuilder:rbac:groups=maas.opendatahub.io,resources=externalmodels/status,verbs=get;update;patch
```
nirrozenbaum: are we planning to update status? (kubebuilder rbac shows we do, IMO not needed atm)

pkg/controller/legacymigration/reconciler_test.go
```diff
@@ -0,0 +1,317 @@
...
+			},
+		},
+	}
+	require.NoError(t, k8sClient.Create(ctx, manual))
+
+	// Create a legacy CR with the same name
+	createLegacyExternalModel(t, "manual-provider", ns, "openai", "api.openai.com", "gpt-4o", "other-key")
+
+	// Wait for reconcile to run
+	time.Sleep(3 * time.Second)
```
nirrozenbaum: can we replace flaky sleep with eventually?

nirrozenbaum (commented):
@yossiovadia the PR is in good shape. 
I left few minor comments and we need to rebase to fix the failing e2e that you already fixed in the other PR.

pkg/controller/legacymigration/reconciler.go
```diff
@@ -0,0 +1,208 @@
...
+
+func isManagedByMigration(labels map[string]string) bool {
+	return labels != nil && labels[labelManagedBy] == managedByValue
+}
+
+func mapProviderToAPIFormat(provider string) string {
+	if provider == "anthropic" {
+		return "messages"
+	}
+	return "openai-chat"
```
nirrozenbaum, line 207: nice

nirrozenbaum (approved):
/lgtm
/approve

nirrozenbaum:
/retest

@yossiovadia pay attention that while the reconciler itself looks great it's still not hooked to controller manager.
IMO we can hook it to plugin resolver, no need for additional plugin (plugin resolver did have previously reconciler watching the old gvk, so it's fine to keep it for deprecation period).

PR #301, opened 2026-06-04
feat: multi-provider API passthrough for external model serving

noyitz:
## Summary

Add **passthrough mode** to the IPP — when the incoming client API format matches the upstream provider's `apiFormat`, request/response translation is skipped entirely. This preserves provider-specific features (prompt caching, extended thinking, beta flags) that are lost during OpenAI↔provider translation.

- Accept `/v1/messages` (Anthropic) and `/v1/responses` (OpenAI Responses API) in addition to `/v1/chat/completions`
- Auto-detect incoming client format from request path
- Skip api-translation when incoming format matches upstream `apiFormat`
- Write both `incoming-api-format` and `api-format` to CycleState for downstream plugin decisions

## Why

- **Claude Code** sends Anthropic Messages format — currently rejected by IPP
- **OpenAI Codex** sends Responses API format — currently rejected by IPP
- Translation is lossy — provider-specific features (prompt caching, thinking, beta flags) have no OpenAI equivalent
- Validated in MaaS dogfood environment with 30 engineers using Claude Code + Codex through the same gateway

## Design

[Design doc](docs/design-multi-provider-passthrough.md) — architecture, data flow, changes required

[Feature refinement](docs/feature-refinement-multi-provider-passthrough.md) — full RHAISTRAT template

## Changes

| File | Change |
|------|--------|
| `pkg/plugins/common/state/state-keys.go` | Add `APIFormatKey` and `IncomingAPIFormatKey` |
| `pkg/plugins/model-provider-resolver/store.go` | Add `apiFormat` to `externalModelInfo` |
| `pkg/plugins/model-provider-resolver/external_model_reconciler.go` | Populate `apiFormat` |
| `pkg/plugins/model-provider-resolver/plugin.go` | Detect incoming format, accept multi-format paths, write to CycleState |
| `pkg/plugins/api-translation/plugin.go` | Skip translation when formats match (passthrough) |
| `pkg/plugins/model-provider-resolver/plugin_test.go` | 5 new tests |

## Dependencies

- **RHAISTRAT-1720** (merged) — `ExternalProviderRef.apiFormat` field
- **MaaS AuthPolicy changes** (not in this PR) — `x-api-key` header support needed for Claude Code auth. Tracked as MaaS team work.
- **BBR Framework SSE fix** (PR #138) — streaming response parsing for non-OpenAI formats

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/plugins/model-provider-resolver/...` passes (10 tests)
- [x] `go test ./pkg/plugins/api-translation/...` passes (all sub-packages)
- [ ] E2E: Anthropic passthrough (client=anthropic, upstream=anthropic)
- [ ] E2E: OpenAI Responses passthrough
- [ ] E2E: Mixed mode (client=openai, upstream=anthropic → translation activates)

## Summary by CodeRabbit

* **New Features**
  * Added multi-API passthrough mode to skip request/response translation when input and output API formats match
  * Enabled API format detection from request paths to drive passthrough decisions

* **Documentation**
  * Added design doc detailing multi-provider passthrough behavior, decision flow, and limitations

* **Tests**
  * Added coverage for passthrough logic and path-based API format detection (unit and e2e)

* **Chores**
  * Updated OpenAI model configuration format for gpt4


yossiovadia:
## OpinAI PR Review

**PR:** #301
**Title:** PR #301
**Author:** @noyitz
**Verdict:** CHANGES_REQUESTED
**Risk:** HIGH
**Base branch:** main

OK, SigV4 isn't in this codebase yet. Let me now finalize my analysis by checking whether the design doc claims match the implementation:

Looking at the design doc's "Changes Required" table vs. the actual diff:

| Component | Claimed Change | Actual Change |
|-----------|---------------|---------------|
| `state-keys.go` | Add `APIFormatKey` and `IncomingAPIFormatKey` | ✅ Done |
| `store.go` | Add `apiFormat` field | ✅ Done |
| `external_model_reconciler.go` | Populate `apiFormat` from CRD (legacy: defaults to provider name) | ⚠️ Only defaults to provider name, never reads from CRD |
| `plugin.go` | Detect incoming format, accept multi-format paths, write to CycleState | ✅ Done, but model mismatch check removed without mention |
| `api-translation/plugin.go` | Skip translation when formats match | ✅ Done |

Now I have all the information for a thorough review.

===PR_VERDICT===
verdict: CHANGES_REQUESTED
risk: HIGH
===END_PR_VERDICT===

--- OPINAI PR REVIEW ---

## Summary

This PR adds **passthrough mode** to the Inference Payload Processor (IPP), allowing clients that speak a provider's native API format (Anthropic `/v1/messages`, OpenAI Responses `/v1/responses`) to bypass the api-translation plugin when the incoming format matches the upstream provider's declared `apiFormat`. The change touches the model-provider-resolver plugin (format detection, multi-path acceptance, CycleState propagation) and the api-translation plugin (skip logic).

## Risk Assessment

**HIGH** — This PR introduces a behavioral regression (model mismatch validation silently removed), has a format mismatch between the CRD schema and the reconciler defaults, accepts paths it cannot handle, and lacks test coverage for the new passthrough logic in the api-translation plugin.

## Previously Identified

No significant code review findings from other reviewers. CodeRabbit skipped the review (draft PR). CI bot only ran group tests.

## New Findings

### 1. **BUG (Blocking): Model mismatch validation silently removed**

The diff removes the model mismatch validation without any mention in the PR description or design doc:

**Removed from `plugin.go`:**
```go
    logger.Error(nil, "model mismatch between request body and ExternalModel", ...)
    return errcommon.Error{Code: errcommon.NotFound, Msg: ...}
}
```

This was a correctness guard ensuring the `model` field in the request body matches the `targetModel` on the resolved ExternalModel CR. Without it, a request with `model: wrong-model` sent to `/llm/claude/v1/chat/completions` would be processed using `claude`'s provider credentials and configuration. The existing E2E test reports (at `test/e2e/reports/3.4/external-model-e2e-report.md`, line 568) confirm this was a tested behavior returning 404.

**In passthrough mode specifically, this is worse:** the body flows through untouched, so the wrong model name would be sent to the upstream provider with the wrong credentials.

This removal is likely unintentional — it's not mentioned anywhere in the PR description, design doc, or commit message.

### 2. **BUG (Blocking): Reconciler never reads `apiFormat` from CRD — hardcodes provider name**

At `external_model_reconciler.go` line 76, the reconciler sets:
```go
apiFormat: provider,
```

The design doc and feature refinement claim "The existing `ExternalProviderRef.apiFormat` field (from [RHAISTRAT-1720](https://redhat.atlassian.net/browse/RHAISTRAT-1720)) already declares the upstream format." However, the reconciler never reads this field from the CRD. It watches the legacy `maas.opendatahub.io/v1alpha1` CRD which has `spec.provider` (flat field), not the new `inference.opendatahub.io/v1alpha1` CRD which has `spec.externalProviderRefs[].apiFormat`.

Furthermore, the sample manifests at `deploy/samples/manifests/anthropic.yaml` use `apiFormat: messages` and `deploy/samples/manifests/openai.yaml` uses `apiFormat: chat-completions` — values that would **never** match the path-detected formats (`"anthropic"`, `"openai"`, `"openai-responses"`). This means passthrough would never activate for the new CRD format.

The comment says "For legacy CRDs, apiFormat defaults to the provider name" — but this is the *only* code path, making the passthrough feature work only for the legacy CRD accidentally (because `provider.Anthropic == "anthropic"` matches the detected format `"anthropic"`).

### 3. **BUG (Medium): `/v1/responses` path accepted but no translator can handle it**

The PR adds acceptance for `/v1/responses` (detected as `"openai-responses"`) in `detectIncomingAPIFormat`. However:
- The reconciler sets `apiFormat = provider` (e.g., `"openai"`)
- `"openai-responses" != "openai"` → passthrough is NOT triggered
- The request falls through to the OpenAI translator, which validates `messages` field (`openai.go` line 47: `if len(messages) == 0 { return errcommon.Error{...} }`)
- The OpenAI Responses API uses `input` and `instructions` fields, not `messages`
- **Result:** All `/v1/responses` requests will be rejected with "messages field is required and must not be empty"

This is a regression from the previous behavior, which rejected `/v1/responses` requests with a clear error ("only /chat/completions input type is supported"). Now the error message is misleading.

### 4. **Design Issue (Medium): Duplicated passthrough logic in ProcessRequest and ProcessResponse**

The passthrough check is copy-pasted in both `ProcessRequest` (line 148-154) and `ProcessResponse` (line 204-208) of `api-translation/plugin.go`:
```go
incomingFormat, _ := framework.ReadCycleStateKey[string](cycleState, state.IncomingAPIFormatKey)
apiFormat, _ := framework.ReadCycleStateKey[string](cycleState, state.APIFormatKey)
if incomingFormat != "" && apiFormat != "" && incomingFormat == apiFormat && incomingFormat != "openai" {
```

This should be extracted to a helper function to ensure consistency and reduce maintenance risk.

### 5. **Concern (Medium): `incomingFormat != "openai"` guard rationale unclear and potentially wrong**

The passthrough condition at `plugin.go` line 150 includes `incomingFormat != "openai"`. This means OpenAI `/v1/chat/completions` requests to an OpenAI provider will always go through translation, even though the OpenAI translator does essentially nothing (returns nil body, rewrites path to `/v1/chat/completions` which it already is).

While this is safe, the exclusion should be documented with a comment explaining *why* openai-to-openai is excluded from passthrough. **ASSUMPTION:** This might be because the OpenAI translator also rewrites the `:path` header, which is needed to strip the model prefix path. If that's the case, the same path-rewriting need would apply to Anthropic passthrough — but in passthrough mode, the path is NOT rewritten, which could cause the upstream provider to receive the full MaaS-prefixed path (e.g., `/llm/claude/v1/messages` instead of `/v1/messages`).

### 6. **Test Coverage Gap (Medium): No passthrough tests in api-translation plugin**

I searched `pkg/plugins/api-translation/plugin_test.go` for `IncomingAPIFormat`, `APIFormat`, `incoming-api-format`, `api-format`, and `passthrough` (in context of the new feature) — none found. The test file at lines 93 and 246 mentions "passthrough" but only for the existing no-provider behavior, not the new format-matching logic.

The new passthrough logic (format comparison, `authorization` removal in passthrough, response skip) in `api-translation/plugin.go` has zero test coverage. This is critical because the passthrough check has subtle conditions (`incomingFormat != "openai"`) that should be validated.

### 7. **Informational: NeMo guardrails interaction with passthrough**

When passthrough mode is active for Anthropic requests, if NeMo request guard is in the plugin chain, it calls `extractMessages(request.Body)` which looks for a top-level `"messages"` key (at `request_guard.go` line 100). Anthropic Messages API bodies DO have a `"messages"` field, so this will work. However, the NeMo response guard (`response_guard.go` line 97) looks for `"choices"` in the response body — Anthropic responses use `"content"` instead. This means NeMo response guardrails silently skip inspection for Anthropic passthrough responses (returning `nil, nil` from `extractAssistantMessages`). This is consistent with the known meta-learning `pattern_nemo_response_guard_message_extraction`, and while not introduced by this PR, the passthrough mode makes this gap more likely to be hit in production.

## Suggestions

1. **Restore the model mismatch check** — or if its removal is intentional, document it in the PR description and add a comment explaining why it's safe to remove.

2. **Read `apiFormat` from the CRD** — either from the new `externalProviderRefs[].apiFormat` or keep the provider-name default with a TODO. Currently the design says one thing and the code does another.

3. **Either implement OpenAI Responses API passthrough properly or reject `/v1/responses` early** — don't accept a path that no translator can handle. A clear "unsupported" error is better than a confusing "messages field is required" error.

4. **Extract the passthrough check into a helper function** — `isPassthrough(cycleState) bool` — used by both `ProcessRequest` and `ProcessResponse`.

5. **Add unit tests for passthrough in api-translation** — test cases needed:
   - Anthropic passthrough (incoming=anthropic, apiFormat=anthropic → skip)
   - OpenAI-to-Anthropic (incoming=openai, apiFormat=anthropic → translate)
   - OpenAI-to-OpenAI (incoming=openai, apiFormat=openai → translate, not passthrough)
   - Passthrough response skip
   - Authorization header removed in passthrough mode

6. **Document the `incomingFormat != "openai"` exclusion** with a comment explaining the reasoning.

## Test Coverage

- **Model-provider-resolver tests:** 5 new tests added, covering Anthropic `/v1/messages`, OpenAI `/v1/responses`, unsupported path, and `detectIncomingAPIFormat`. Good coverage for the format detection logic.
- **Api-translation tests:** No new tests for the passthrough logic. This is a significant gap — the core feature (skip translation when formats match) is untested in the plugin that implements it.
- **Reconciler tests:** No tests for the `apiFormat` field population. The existing reconciler test (if any) doesn't verify `apiFormat` is set correctly.
- **Missing edge case tests:** What happens when only one of `IncomingAPIFormatKey`/`APIFormatKey` is set? The code handles this (both must be non-empty), but it's not tested.

## Verdict

**CHANGES_REQUESTED** — Three blocking/high-severity issues:
1. Silent removal of model mismatch validation creates a correctness regression
2. Reconciler never reads `apiFormat` from CRD, making the feature only work by accident for legacy CRDs
3. `/v1/responses` path accepted but will always fail at translation with a misleading error

These need to be addressed before merge. The passthrough concept is sound, but the implementation has gaps between the design document's claims and the actual code.

--- END PR REVIEW ---

--- OPINAI SUGGESTED_QUESTIONS ---
[
  "In `external_model_reconciler.go`, you set `apiFormat: provider` as a default. The new CRD schema (`inference.opendatahub.io/v1alpha1`) has `externalProviderRefs[].apiFormat` with values like `messages` and `chat-completions` per the sample manifests — should the reconciler read this field? If so, how do you plan to handle the format-name mismatch between CRD values (`messages`) and path-detected values (`anthropic`)?",
  "The model mismatch check (`externalModelInfo.targetModel != model`) was removed without mention in the PR description. Was this intentional? In passthrough mode, the request body flows to the upstream provider unchanged — if a client sends `model: wrong-model` with valid credentials for a different model, the upstream provider would receive the wrong model name with those credentials.",
  "When a client sends `/v1/responses` (OpenAI Responses API), the incoming format is `openai-responses` but the apiFormat defaults to `openai` — so passthrough doesn't trigger, and the OpenAI translator rejects the request because it requires a `messages` field. What's the intended behavior for Responses API requests in this PR? Should they only work in passthrough, or do you plan a separate translator?",
  "In passthrough mode, the api-translation plugin still calls `request.RemoveHeader('authorization')` but does NOT rewrite the `:path` header to strip the model prefix (e.g., `/llm/claude/v1/messages` → `/v1/messages`). The normal translation path rewrites `:path` via the translator. How is the path-prefix stripping handled for passthrough requests — is there an upstream URLRewrite filter, or does the provider receive the full MaaS-prefixed path?",
  "The `incomingFormat != 'openai'` condition in the passthrough check excludes OpenAI-to-OpenAI from passthrough. Is this because the OpenAI translator's `:path` rewrite is needed even for OpenAI targets? If so, wouldn't the same path-rewriting concern apply to Anthropic passthrough where the path is also prefixed?"
]
--- END SUGGESTED_QUESTIONS ---

yossiovadia:
## OpinAI PR Review

**PR:** #301
**Title:** PR #301
**Author:** @noyitz
**Verdict:** CHANGES_REQUESTED
**Risk:** HIGH
**Base branch:** main

Now I have a complete picture. Let me deliver the review.

===PR_VERDICT===
verdict: CHANGES_REQUESTED
risk: HIGH
===END_PR_VERDICT===

--- OPINAI PR REVIEW ---

## Summary

This PR adds **passthrough mode** to the IPP: when a client sends a request in the same API format as the upstream provider (e.g., Claude Code sending Anthropic Messages to an Anthropic backend), request/response translation is skipped. The change introduces format detection from request paths (`/v1/messages`, `/v1/responses`, `/v1/embeddings`), writes input/output format keys to CycleState, and adds an `isPassthrough()` check in the api-translation plugin.

## Risk Assessment

**HIGH** — The reconciler comment claims to support a new CRD format but doesn't actually read it, the `apiFormat` naming convention has a hard-coded mismatch with sample manifests, `CredsRefNamespace` test coverage was silently dropped, and the `isPassthrough` function has no unit tests in the api-translation package.

## Previously Identified

- @yossiovadia (previous OpinAI review) already flagged the PR as CHANGES_REQUESTED with HIGH risk and identified the design doc vs implementation divergence pattern.

## New Findings

### 1. **BUG (Blocking): Reconciler never reads `apiFormat` from the new CRD — comment claims it does**

`pkg/plugins/model-provider-resolver/external_model_reconciler.go` lines 71-72 state:
> "For legacy CRDs (maas.opendatahub.io), apiFormat is derived from the provider name. The new CRD (inference.opendatahub.io) has an explicit apiFormat field on ExternalProviderRef."

And line 91:
> "For the new CRD (inference.opendatahub.io), the apiFormat field is read directly."

**But neither of these claims is implemented.** The reconciler:
1. Only watches `maas.opendatahub.io/v1alpha1/ExternalModel` (line 34)
2. Only reads `spec.provider`, `spec.targetModel`, and `spec.credentialRef.name` (lines 67-69)
3. **Always** calls `providerToAPIFormat(provider)` — never reads `apiFormat` from the CRD (line 76)

Meanwhile, the new CRD at `config/crd/bases/inference.opendatahub.io_externalmodels.yaml` has `apiFormat` as a **required field** under `externalProviderRefs[].apiFormat`, but the reconciler doesn't read it at all. The code tells the story that future support is planned, but the comments are misleading — they claim functionality that doesn't exist.

### 2. **BUG (Blocking): `apiFormat` naming mismatch between sample manifests and code**

`deploy/samples/manifests/openai.yaml` line 31 uses:
```yaml
apiFormat: chat-completions
```

But `providerToAPIFormat("openai")` in `external_model_reconciler.go` line 97 returns `"openai-chat"`. The design doc table also uses `"openai-chat"`. There's no code path that maps `"chat-completions"` → `"openai-chat"`. If a user creates an ExternalModel with the new CRD using the sample manifest's format, and the reconciler eventually reads the CRD field directly, passthrough comparison will fail silently — `"chat-completions" != "openai-chat"`.

### 3. **BUG (Medium): `TestProcessRequest_ModelResolved` silently dropped `CredsRefNamespace` assertion**

The diff shows lines 71-73 of the original test verified `CredsRefNamespace`:
```go
actualCredsNamespace, err := framework.ReadCycleStateKey[string](cs, state.CredsRefNamespace)
require.NoError(t, err)
require.Equal(t, extNS, actualCredsNamespace)
```

This was replaced with the `OutputAPIFormatKey` assertion. The `CredsRefNamespace` assertion should be **added back** alongside the new assertions, not replaced. This loses regression coverage for a critical CycleState key used by `apikey-injection`.

### 4. **Missing test coverage: `isPassthrough()` has no tests in `api-translation/plugin_test.go`**

I searched `pkg/plugins/api-translation/plugin_test.go` for `isPassthrough`, `InputAPIFormatKey`, and `OutputAPIFormatKey` — no matches found. The `isPassthrough()` function (lines 245-256 of `plugin.go`) has three important branches:
- Empty format keys → false
- Mismatched formats → false  
- Matching formats but `openai-chat` excluded → false
- Matching non-`openai-chat` formats → true

None of these are tested in the api-translation tests. The model-provider-resolver tests verify CycleState is written correctly, but no test verifies that `ProcessRequest`/`ProcessResponse` actually skip translation when formats match.

### 5. **Design gap (Medium): Passthrough skips `:path` rewriting — upstream receives MaaS-prefixed path**

In passthrough mode (`plugin.go` lines 148-151), `ProcessRequest` only calls `request.RemoveHeader("authorization")` and returns. The existing translation code path for Anthropic (`anthropic.go`) rewrites `:path` to `/v1/messages`. In passthrough mode, the upstream will receive the full MaaS-prefixed path like `/llm/claude/v1/messages`.

The design doc (line 168) acknowledges this and says "path rewriting is handled by the HTTPRoute URLRewrite filter." However, this is a **MaaS dependency** listed under "Required for Production" (design doc line 152). This means this PR alone produces broken paths for any deployment without the corresponding MaaS changes. This should be documented more prominently or gated.

### 6. **Minor: `detectInputAPIFormat` matches overly broad suffixes**

`plugin.go` line 156: `strings.HasSuffix(path, "/messages")` would match any path ending in `/messages`, such as `/v2/some-custom/messages` or even `/error-messages`. While the path is constructed from a sanitized URL, this is fragile. Similarly, `/responses` and `/embeddings` are common words that could appear in other paths.

### 7. **Minor: PR description references non-existent document**

The PR description links to `docs/feature-refinement-multi-provider-passthrough.md` but this file doesn't exist in the diff or in the `docs/` directory (which only contains `design-multi-provider-passthrough.md` and `providers/`).

## Suggestions

1. **Remove or qualify the misleading comments** in `external_model_reconciler.go` lines 71-72 and 91. They should say "TODO: read apiFormat from new CRD when reconciler supports inference.opendatahub.io" rather than implying it already works.

2. **Standardize `apiFormat` naming**: Decide on a canonical set of format identifiers and use them consistently in code, CRDs, sample manifests, and design docs. Currently: code uses `"openai-chat"`, sample manifest uses `"chat-completions"`, CRD type doc says `"openai"`.

3. **Add `isPassthrough()` unit tests** in `pkg/plugins/api-translation/plugin_test.go` — at minimum test passthrough activation (Anthropic messages match), passthrough skipping (format mismatch), and the `openai-chat` exclusion.

4. **Restore the `CredsRefNamespace` assertion** in `TestProcessRequest_ModelResolved`.

5. Consider adding a `TestProcessRequest_AnthropicPassthrough_FullPipeline` integration-style test that exercises both model-provider-resolver and api-translation together to verify the end-to-end passthrough flow.

## Test Coverage

**Model-provider-resolver tests** are good: 5 new tests cover Anthropic messages path, OpenAI responses path, unsupported path rejection, format detection, and provider-to-format mapping.

**Api-translation tests** have a gap: the `isPassthrough()` function introduced in this PR has **zero test coverage** in its own package. I searched `pkg/plugins/api-translation/plugin_test.go` for any reference to `isPassthrough`, `InputAPIFormatKey`, or `OutputAPIFormatKey` and found none.

The `CredsRefNamespace` assertion was dropped from `TestProcessRequest_ModelResolved`, reducing regression coverage for downstream plugins.

## Verdict

**CHANGES_REQUESTED.** The misleading comments about new CRD support, the `apiFormat` naming inconsistency with sample manifests, the dropped test assertion, and the missing `isPassthrough()` test coverage should be addressed before merge. The core design is sound and well-structured, but these issues create real risk of silent failures in production.

--- END PR REVIEW ---

--- OPINAI SUGGESTED_QUESTIONS ---
[
  "What happens when a user deploys an `inference.opendatahub.io` ExternalModel CR with `apiFormat: chat-completions` (from your sample manifest) — will passthrough ever activate, given the reconciler derives format from provider name and returns `openai-chat`?",
  "In passthrough mode, the upstream receives the MaaS-prefixed path (e.g., `/llm/claude/v1/messages`) since `:path` rewriting is skipped. Is there a specific HTTPRoute URLRewrite filter that's guaranteed to be deployed alongside this PR, or can this break existing Anthropic deployments?",
  "The `isPassthrough` function excludes `openai-chat` because the OpenAI translator rewrites `:path`. But the Anthropic translator also rewrites `:path` to `/v1/messages` — why is it safe to skip that rewrite in passthrough mode? Is the assumption that the path is already correct since the client sent to `/v1/messages`?",
  "The `detectInputAPIFormat` function maps `/embeddings` to a format identifier, but no translator or passthrough logic handles embeddings yet. Should requests to `/v1/embeddings` for external models succeed (silently passing through) or fail explicitly until embeddings support is implemented?",
  "In the `TestProcessRequest_ModelResolved` test, the `CredsRefNamespace` assertion was replaced with `OutputAPIFormatKey`. Was this intentional, or should both assertions coexist to maintain regression coverage for the apikey-injection plugin's dependency on `CredsRefNamespace`?"
]
--- END SUGGESTED_QUESTIONS ---

yossiovadia (approved):
/LGTM

PR #304, opened 2026-06-08
feat: add apiFormat to store + CycleState, model reconciler unit tests

yossiovadia:
## Summary

Follow-up to #286 per Nir's review comments:

- **apiFormat in store + CycleState** — store `ref.APIFormat` in `externalModelInfo`, write `APIFormatKey` to CycleState so api-translation can select the right translator
- **Always write config to CycleState** — removed `len > 0` guard, downstream plugins don't need nil checks
- **Initialize empty config map** — provider store initializes `map[string]string{}` when `spec.config` is nil
- **Model reconciler unit tests** — 7 tests covering: happy path, delete, provider not available (requeue), multi-ref fallback, auth override, config merge, and mergeConfig edge cases (nil maps, overlapping keys, copy safety)

Coverage for `model-provider-resolver` went from 48% → 75%.

**Not in this PR** (separate follow-ups):
- Legacy migration wiring into plugin.go (needs #299 merged first)
- Weight-based provider selection (new scope from Nir)

## Test plan

- [x] `make test-unit` — all tests pass (75% coverage)
- [x] `make lint` — 0 issues
- [x] 7 new model reconciler tests all pass
- [x] mergeConfig tests verify copy safety (no shared map mutation)

## Summary by CodeRabbit

* **New Features**
  * API format information is now resolved and propagated through the model provider reconciliation process.

* **Improvements**
  * Provider configuration is now properly normalized and validated during reconciliation.

* **Tests**
  * Added comprehensive test coverage for model and provider reconciliation scenarios including fallback and override behaviors.

### DIFF
--- a/pkg/plugins/model-provider-resolver/store.go
+++ b/pkg/plugins/model-provider-resolver/store.go
@@ -35,6 +35,7 @@ type providerInfo struct {
 type externalModelInfo struct {
 	provider        string
 	targetModel     string
+	apiFormat       string
 	secretName      string
 	secretNamespace string
 	config          map[string]string
--- a/pkg/plugins/common/state/state-keys.go
+++ b/pkg/plugins/common/state/state-keys.go
@@ -23,4 +23,5 @@ const (
 	CredsRefName      = "credential-ref-name"
 	CredsRefNamespace = "credential-ref-namespace"
 	ModelConfigKey    = "model-config"
+	APIFormatKey      = "api-format"
 )
--- a/pkg/plugins/model-provider-resolver/external_model_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_model_reconciler.go
@@ -92,6 +92,7 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1al
 	return &externalModelInfo{
 		provider:        providerInfo.provider,
 		targetModel:     ref.TargetModel,
+		apiFormat:       ref.APIFormat,
 		secretName:      secretName,
 		secretNamespace: secretNamespace,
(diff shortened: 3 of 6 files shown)


pkg/plugins/model-provider-resolver/store.go
```diff
@@ -35,6 +35,7 @@ type providerInfo struct {
 type externalModelInfo struct {
 	provider        string
 	targetModel     string
+	apiFormat       string
 	secretName      string
 	secretNamespace string
 	config          map[string]string
```
nirrozenbaum, line 41: that's a good progress. 
next we would need to add multi provider support and that should be reflected here.
all these fields should be stored per provider.
and then external model info should be a per provider info + weight.
on a new request the resolver plugin would randomly select the provider (based on the weights) and then inject to cycle state the right values.

can be done in a follow up.

nirrozenbaum (approved):
/lgtm

nice. can be merged as incremental progress.

we still have a list of open items to handle in follow up:
- create new PR to hook legacy controller to plugin resolver (register the controller to the manager). no need for a new plugin just hook to the existing plugin (should be 5 lines).
- add weight to provider ref. if no weight exist default to weight = 1. should be stored in store and used by plugin resolver to select the provider based on the weight.
- update the code in the reconciler to support multiple provider refs, adding all entries to the store (including weight)
- document somewhere the available api format and what are they mapped to (e.g., openai-chat is mapped to chat completions).
- update deploy/samples manifests to include a full example with all the fields including config, api format, weight, etc (e.g., have a vertex example).
- go over the new CRDs godoc comments, some of them are wrong.

PR #306, opened 2026-06-09
ci(pr-size-labeler): update comment in place instead of duplicating

yehuditkerido:
## Description

Update the PR size labeler workflow to update the existing "Large PR detected" comment instead of creating a new comment on every push. Uses `actions/github-script` to find the comment by a hidden marker and update it in place.

## How Has This Been Tested?

Tested on a fork by pushing multiple times to a large PR and verifying the comment was updated rather than duplicated.

## Summary by CodeRabbit

* **Chores**
  * Improved automation for labeling and holding large pull requests, including safer cleanup of prior marker comments to prevent duplicates and consistent application/removal of the hold label as needed.

### DIFF
--- a/.github/workflows/pr-size-labeler.yml
+++ b/.github/workflows/pr-size-labeler.yml
@@ -33,19 +33,42 @@ jobs:
               "1000": "XXL"
             }
 
-      - name: Hold large PRs
-        if: contains(steps.size.outputs.sizeLabel, 'XL')
+      - name: Handle large PR warning
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          gh pr edit ${{ github.event.pull_request.number }} \
-            --repo ${{ github.repository }} \
-            --add-label "do-not-merge/hold"
-          
-          gh pr comment ${{ github.event.pull_request.number }} \
-            --repo ${{ github.repository }} \
-            --body "⚠️ **Large PR detected**
+          SIZE_LABEL: ${{ steps.size.outputs.sizeLabel }}
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const issue_number = context.issue.number;
+            const marker = '<!-- pr-size-labeler-comment -->';
+            const isLarge = process.env.SIZE_LABEL.includes('XL');
 
-          Your PR is large. Please consider breaking it into multiple PRs.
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-size-labeler.yml
```diff
@@ -35,17 +35,28 @@ jobs:
...
+              '',
+              'The `do-not-merge/hold` label has been added and can be removed by the reviewers based on their judgement.'
+            ].join('\n');
 
-          Your PR is large. Please consider breaking it into multiple PRs.
+            await github.rest.issues.addLabels({ owner, repo, issue_number, labels: ['do-not-merge/hold'] });
 
-          The \`do-not-merge/hold\` label has been added and can be removed by the reviewers based on their judgement."
+            const { data: comments } = await github.rest.issues.listComments({ owner, repo, issue_number });
+            const existing = comments.find(c => c.user.type === 'Bot' && c.body.includes(marker));
```
nirrozenbaum: why do we filter by Bot? isn't the marker enough?

nirrozenbaum (commented):
not sure I understood what `updateComment` does. the comment is static (always the same comment) so why do we need the update? 

generally speaking - I'd say that when a push is done, it's probably better to delete the previous comment (if exist) and create a new one, so the `large PR detected` comment is last in the list (otherwise if comment was added long time ago, the user probably miss any update). 

also it would be good to add the other way around - if PR was big and got automatic hold with the marker comment, then it was updated and reduced the size - we can auto clean the marker comment and the `hold` label.

thoughts?

.github/workflows/pr-size-labeler.yml
yehuditkerido: You're right, the marker is enough. Removed the Bot filter.

yehuditkerido:
> not sure I understood what `updateComment` does. the comment is static (always the same comment) so why do we need the update?
> 
> generally speaking - I'd say that when a push is done, it's probably better to delete the previous comment (if exist) and create a new one, so the `large PR detected` comment is last in the list (otherwise if comment was added long time ago, the user probably miss any update).
> 
> also it would be good to add the other way around - if PR was big and got automatic hold with the marker comment, then it was updated and reduced the size - we can auto clean the marker comment and the `hold` label.
> 
> thoughts?

Agree. Updated to:

1. Always delete old comment + remove hold label first (clean slate)
2. Then create new ones only if PR is XL+

Pushing the update now.

nirrozenbaum:
going to close this PR, cause we're removing the auto labeling in PR #341

PR #308, opened 2026-06-09
feat: weight-based multi-provider selection

yossiovadia:
## Summary

Add weight-based traffic splitting across multiple provider refs, per Nir's follow-up items from #304.

**CRD change:**
- New `weight` field on `ExternalProviderRef` (optional, default 1, range 1-100)

**Store restructure:**
- `externalModelInfo` now holds `[]resolvedProviderRef` — all resolved refs with weights
- Reconciler resolves ALL available provider refs (not just first)
- Partially available refs are stored (unavailable ones skipped with log)

**Plugin resolver:**
- `selectByWeight` picks a provider using weighted random selection
- Single ref: no randomness, returns directly
- Multiple refs: weighted random proportional to weights

**Example:**
```yaml
externalProviderRefs:
  - ref: { name: openai-provider }
    targetModel: gpt-4o
    apiFormat: openai-chat
    weight: 80          # 80% traffic
  - ref: { name: azure-provider }
    targetModel: gpt-4o
    apiFormat: openai-chat
    weight: 20          # 20% traffic
```

## Size note

~80 lines of production code. The rest is generated files (deepcopy + CRD YAML, 23 lines) and test updates/additions (~186 lines).

## Test plan

- [x] `make test-unit` — all tests pass (77.3% coverage)
- [x] `make lint` — 0 issues
- [x] `make verify-codegen` — generated files up to date
- [x] TestSelectByWeight_Distribution — verifies 80/20 split over 1000 iterations
- [x] TestSelectByWeight_EqualWeights — verifies even distribution
- [x] TestModelReconciler_WeightFromCRD — verifies weight read from CRD
- [x] TestModelReconciler_WeightDefaultsToOne — verifies nil weight → 1
- [x] TestModelReconciler_MultiRefAllResolved — both refs stored
- [x] TestModelReconciler_MultiRefPartialAvailability — only available ref stored


pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -56,24 +56,30 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-		info, found := r.resolveRef(req.Namespace, ref)
+		rr, found := r.resolveRef(req.Namespace, ref)
 		if !found {
+			logger.Info("ExternalProvider not yet available, skipping ref", "provider", ref.Ref.Name)
 			continue
 		}
-		r.store.addOrUpdateModel(req.NamespacedName, info)
-		logger.Info("updated model store", "provider", info.provider, "targetModel", info.targetModel)
-		return ctrl.Result{}, nil
+		resolved = append(resolved, *rr)
```
nirrozenbaum, line 67: can we use descriptive names?
```suggestion
		resolvedRef, found := r.resolveRef(req.Namespace, ref)
if !found {
			logger.Info("ExternalProvider not yet available, skipping ref", "provider", ref.Ref.Name)
continue
}
		resolved = append(resolved, *resolvedRef)
```

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -56,24 +56,30 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
-	// Use the first ref whose provider is available in the store.
+	// Resolve all refs whose providers are available in the store.
+	var resolved []resolvedProviderRef
```
nirrozenbaum, line 60: why not slice of pointers?

```suggestion
	var resolved []*resolvedProviderRef
```

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -56,24 +56,30 @@ func (r *externalModelReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	r.store.addOrUpdateModel(req.NamespacedName, &externalModelInfo{refs: resolved})
+	logger.Info("updated model store", "resolvedRefs", len(resolved))
+	return ctrl.Result{}, nil
 }
 
-// resolveRef resolves a single ExternalProviderRef to model info.
+// resolveRef resolves a single ExternalProviderRef to provider info.
 // Returns (nil, false) if the provider is not yet available in the store.
-func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1alpha1.ExternalProviderRef) (*externalModelInfo, bool) {
+func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1alpha1.ExternalProviderRef) (*resolvedProviderRef, bool) {
```
nirrozenbaum, line 82: why not pass in ref as pointer?
each call will copy over the whole object

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -89,13 +95,19 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1al
 		secretNamespace = namespace
 	}
 
-	return &externalModelInfo{
+	weight := 1
+	if ref.Weight != nil {
+		weight = *ref.Weight
+	}
```
nirrozenbaum, line 101: this is fine to keep, but just out of curiosity - 
I saw the CRD has default value weight = 1. is it even possible to have weight == nil?

pkg/plugins/model-provider-resolver/external_model_reconciler_test.go
```diff
@@ -49,6 +49,8 @@ func (m *mockModelReader) List(_ context.Context, _ client.ObjectList, _ ...clie
 	return nil
 }
 
+func intPtr(v int) *int { return &v }
```
nirrozenbaum, line 52: can we replace this and use instead `k8s.io/utils/ptr`? (import and use it)

pkg/plugins/model-provider-resolver/external_model_reconciler_test.go
```diff
@@ -151,20 +188,17 @@ func TestModelReconciler_MultiRefFallback(t *testing.T) {
 
 	info, found := store.getModel(key)
 	require.True(t, found)
-	assert.Equal(t, "anthropic", info.provider)
-	assert.Equal(t, "claude-sonnet", info.targetModel)
-	assert.Equal(t, "messages", info.apiFormat)
-	assert.Equal(t, "anthropic-key", info.secretName)
```
nirrozenbaum, line 157: what about these assertions that were removed?
we can easily keep them by using info.refs[0].targetModel, etc.

pkg/plugins/model-provider-resolver/model_store_test.go
```diff
@@ -29,19 +29,21 @@ func TestModelStore_AddAndGetExternalModel(t *testing.T) {
 	store := newInfoStore()
 	key := types.NamespacedName{Namespace: "ns", Name: "external-model"}
 
-	store.addOrUpdateModel(key, &externalModelInfo{provider: provider.Anthropic})
```
nirrozenbaum, line 32: we merged provider store with model store, but the tests are still in separate files (store_test + model_store_test). can we merge the test files to store_test.go?

not needed for this PR can be done in a dedicated PR.

noyitz (approved):
LGTM — clean design, solid tests.

One thing worth considering (not blocking): if two refs have different `targetModel` values (e.g., ref1: openai/gpt-4o, ref2: anthropic/claude-sonnet), the model mismatch check in `ProcessRequest` runs against the randomly-selected ref. A client sending `"model": "gpt-4o"` would get intermittent 404s whenever `selectByWeight` picks the anthropic ref. This is a misconfiguration scenario, but a warning log during reconciliation when refs have mixed targetModels would help operators catch it early. Fine as a follow-up.

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -153,24 +154,47 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 		return errcommon.Error{Code: errcommon.BadRequest, Msg: "only /chat/completions input type is supported"}
 	}
 
-	// model in request body must match the ExternalModel's targetModel
-	if externalModelInfo.targetModel != model {
-		logger.Error(nil, "model mismatch between request body and ExternalModel", "requestModel", model, "externalModel", externalModelInfo.targetModel)
```
nirrozenbaum, line 158: this is a bug. 
we should match the request on the model name not on the target model name (each provider might have a different targetModel name).

for now we're using metadata.name. 
if we later want to avoid the char limitations of metadata.name, we can add a dedicated field on the spec. 

but this should be restored and checked BEFORE selecting by weight.
and we should compare `externalModelInfo.metadata.name != model` (I assume we should add a dedicated field in spec in a follow up, not needed in this PR. it will require updating the samples as well)

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -153,24 +154,47 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
...
+	// select a provider ref based on weights
+	ref := selectByWeight(modelInfo.refs)
+
+	// model in request body must match the selected ref's targetModel
+	if ref.targetModel != model {
+		logger.Error(nil, "model mismatch between request body and ExternalModel", "requestModel", model, "externalModel", ref.targetModel)
```
nirrozenbaum, line 162: as noted in the comment above, this should happen before random by weight

nirrozenbaum (commented):
was just finishing my review and then it got merged :) 
there are few things to fix here, not big but still needed.

since this one is already merged let's do a follow up to fix these points.

PR #309, opened 2026-06-09
docs: update apiFormat naming, sample manifests, and godoc

yossiovadia:
## Summary

Per Nir's follow-up list from #304:

- **apiFormat godoc** — updated to document `openai-chat` and `messages` as the supported values (was listing stale names like `openai`, `anthropic`, `bedrock-openai`)
- **Sample manifests fixed** — added `auth.type: simple` (required after #290), corrected apiFormat values
- **New Vertex AI sample** — includes provider config (`project`, `location`) and model config override (`endpoint`)
- **New multi-provider weighted sample** — 80/20 traffic split between OpenAI and Azure OpenAI
- **CRD YAML regenerated** for updated godoc

## Test plan

- [x] `go build ./...` — compiles
- [x] `make lint` — 0 issues
- [x] `make verify-codegen` — generated files up to date

## Summary by CodeRabbit

* **New Features**
  * Added sample configuration for multi-provider inference with weighted traffic splitting between OpenAI and Azure providers
  * Added Vertex AI provider sample configuration with Gemini model support

* **Documentation**
  * Clarified supported API translation formats ("openai-chat" and "messages") in external provider documentation

PR #307, opened 2026-06-09
feat: wire legacy migration controller into plugin resolver

yossiovadia:
## Summary

Register the legacy migration reconciler (from #299) in `NewModelProviderResolver` to watch `maas.opendatahub.io` ExternalModel CRs. This completes the legacy support flow:

```
Old CR (maas.opendatahub.io) → migration reconciler → creates new CRs (inference.opendatahub.io)
  → typed store reconcilers populate store
  → ProcessRequest resolves model → CycleState
```

Per Nir's comment on #299: "hook it to plugin resolver, no need for additional plugin."

## Test plan

- [x] `make test-unit` — all tests pass
- [x] `make lint` — 0 issues
- [x] 12 lines added, single file change

## Summary by CodeRabbit

* **New Features**
  * Added support for legacy resource migration and monitoring alongside standard provider resources.

PR #310, opened 2026-06-09
feat: wire controller reconcilers into plugin factory

yossiovadia:
## Summary

Wire the controller reconcilers from `pkg/controller/` (merged in #183) into the model-provider-resolver plugin factory. This is the step that makes the CRDs fully operational — creating an ExternalProvider or ExternalModel CR now automatically creates the networking resources.

**Flow after this PR:**
```
ExternalProvider CR → controller creates Service + ServiceEntry + DestinationRule
ExternalModel CR → controller creates HTTPRoute (with gateway parent ref)
```

**Changes:**
- `resolverConfig` struct for gateway name/namespace/timeout from plugin JSON config
- Gateway API scheme registration in factory
- `externalprovider.Reconciler` registered with `Owns(&corev1.Service{})` and managed-by predicate
- `externalmodel.Reconciler` registered with `Owns(&gatewayapiv1.HTTPRoute{})` and cross-watch on ExternalProvider

## Size note

60 lines added, single file change. All production code, no tests needed (controller reconcilers have their own envtest suites in `pkg/controller/`).

## Test plan

- [x] `make test-unit` — all tests pass
- [x] `make lint` — 0 issues

## Summary by CodeRabbit

* **New Features**
  * Enabled additional controller wiring in the main runner to support inference provider resources and gateway-backed model resources.
  * Added dedicated reconciliation setup so model resources automatically refresh in response to related provider changes.
* **Tests**
  * Updated end-to-end simulator behavior to use a derived simulator hostname when creating provider resources.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,6 +28,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"github.com/llm-d/llm-d-inference-payload-processor/cmd/runner"
 
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/controllers"
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins"
 )
 
@@ -37,7 +38,13 @@ func main() {
 
 	if err := runner.NewRunner().
 		WithExecutableName("ai-gateway-payload-processing").
-		// WithCustomCollectors(...). // THIS should be used for custom metrics exposed by our plugins
+		WithCustomControllers(
+			controllers.ProviderController(),
+			controllers.ModelController(
+				os.Getenv("GATEWAY_NAME"),
+				os.Getenv("GATEWAY_NAMESPACE"),
+			),
+		).
 		Run(ctrl.SetupSignalHandler()); err != nil {
 		os.Exit(1)
 	}
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -32,8 +32,9 @@ var (
 	gatewayNs      string
 	gatewayName    string
 	gatewaySvcName string
-	simulatorEP    string
-	curlTimeout    = 30 * time.Second
+	simulatorEP   string
+	simulatorFQDN string
+	curlTimeout   = 30 * time.Second
 )
 
 func TestE2E(t *testing.T) {
@@ -47,6 +48,7 @@ var _ = ginkgo.BeforeSuite(func() {
 	gatewayName = envOr("E2E_GATEWAY_NAME", defaultGatewayName)
 	gatewaySvcName = envOr("E2E_GATEWAY_SVC_NAME", gatewayName+"-istio")
 	simulatorEP = envOr("E2E_SIMULATOR_ENDPOINT", defaultSimulatorEndpoint)
+	simulatorFQDN = strings.ReplaceAll(simulatorEP, ".", "-") + ".sslip.io"
 
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		clientcmd.NewDefaultClientConfigLoadingRules(), nil,
@@ -66,38 +68,6 @@ var _ = ginkgo.AfterSuite(func() {
 func setupInfra() {
 	ginkgo.By("Creating test namespace")
 	createNamespace(nsName)
-
-	ginkgo.By("Creating simulator ServiceEntry and DestinationRule")
-	kubectlApplyLiteral(fmt.Sprintf(`
-apiVersion: networking.istio.io/v1
-kind: ServiceEntry
-metadata:
-  name: e2e-simulator
-  namespace: %s
-spec:
-  hosts:
-  - e2e-simulator.external
-  location: MESH_EXTERNAL
-  ports:
-  - number: 443
-    name: https
-    protocol: HTTPS
-  resolution: STATIC
-  endpoints:
-  - address: %s
----
-apiVersion: networking.istio.io/v1
-kind: DestinationRule
-metadata:
-  name: e2e-simulator
-  namespace: %s
-spec:
-  host: e2e-simulator.external
-  trafficPolicy:
-    tls:
-      mode: SIMPLE
-      insecureSkipVerify: true
-`, nsName, simulatorEP, nsName))
 
 	ginkgo.By("Creating curl client pod")
 	kubectlApplyLiteral(fmt.Sprintf(`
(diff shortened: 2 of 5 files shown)


yossiovadia:
## Local Integration Test Results

Tested all 4 PRs (#307, #308, #309, #310) merged together on a local Kind cluster with real traffic against the llm-katan simulator.

| Test | Result |
|------|--------|
| New ExternalProvider CR → auto-creates Service + ServiceEntry + DestinationRule | ✅ |
| New ExternalModel CR → auto-creates HTTPRoute | ✅ |
| Send request → model resolved → api-translation → response 200 OK | ✅ |
| Legacy `maas.opendatahub.io` CR → migration creates new CRs → auto-creates resources → 200 OK | ✅ |
| Weight field accepted by CRD | ✅ |
| apiFormat stored in CycleState | ✅ |
| ExternalProvider status: Ready | ✅ |
| ExternalModel status: Ready | ✅ |

**E2E CI failure:** Expected — the controller wiring registers watchers for Services and HTTPRoutes, which require WRITE RBAC (create/update/delete). The current Helm chart only has READ permissions. Without WRITE RBAC, controller-runtime enters a retry loop that blocks the ext-proc server.

**Fix:** The Helm RBAC PR (next after this one) will add the controller WRITE permissions. For this PR, the E2E failure is not a code issue — it's a missing RBAC configuration that the next PR adds.

nirrozenbaum:
@yossiovadia this is an important PR that connect the dots, but I think it should be done differently. 
it actually reveals a gap we have in the IPP framework - when we want to have a control plane controllers, we must create "dummy" plugins for them (or register it to existing plugin). 

this seems like a wrong pattern. 
we need to add a mechanism to register controllers in IPP framework (in runner.go, e.g., add `WithCustomControllers` function).

then we would need to register these two controllers using it and also move the legacy migration controller to that pattern. 

controllers registered within plugins are "Data Plane" related, e.g., controller that watches secrets to fill in the secret store that is used when a request comes in.
let's work on the right solution.

holding to make sure this is doesn't get merged by mistake.
/hold

nirrozenbaum:
> @yossiovadia this is an important PR that connect the dots, but I think it should be done differently. it actually reveals a gap we have in the IPP framework - when we want to have a control plane controllers, we must create "dummy" plugins for them (or register it to existing plugin).
> 
> this seems like a wrong pattern. we need to add a mechanism to register controllers in IPP framework (in runner.go, e.g., add `WithCustomControllers` function).
> 
> then we would need to register these two controllers using it and also move the legacy migration controller to that pattern.
> 
> controllers registered within plugins are "Data Plane" related, e.g., controller that watches secrets to fill in the secret store that is used when a request comes in. let's work on the right solution.
> 
> holding to make sure this is doesn't get merged by mistake. /hold

@yossiovadia thinking about this again. I don't want to block this functionality until we do the required change in IPP framework inside llm-d (we will do it, just do want to block on it). 
so let's go ahead and rebase this PR to get it to the finish line. in parallel we can work on extending the llm-d framework and as soon as that is ready we can do the switch.

/unhold

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -111,6 +131,40 @@ func NewModelProviderResolver(reconcilerBuilder func() *builder.Builder, k8sClie
...
+		Complete(ctrlProviderReconciler); err != nil {
+		return nil, fmt.Errorf("failed to register ExternalProvider controller: %w", err)
+	}
+
+	ctrlModelReconciler := &externalmodel.Reconciler{
+		Client:           k8sClient,
+		Scheme:           k8sClient.Scheme(),
+		GatewayName:      config.GatewayName,
+		GatewayNamespace: config.GatewayNamespace,
+		RouteTimeout:     config.RouteTimeout,
```
nirrozenbaum: where are these configured? omitempty on the config json will result in empty values here, won't it?

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -48,20 +54,34 @@ const (
...
+	GatewayNamespace string `json:"gatewayNamespace,omitempty"`
+	RouteTimeout     string `json:"routeTimeout,omitempty"`
+}
+
 // ModelProviderResolverFactory defines the factory function for ModelProviderResolverPlugin.
-func ModelProviderResolverFactory(name string, _ json.RawMessage, handle framework.Handle) (framework.BBRPlugin, error) {
-	plugin, err := NewModelProviderResolver(handle.ReconcilerBuilder, handle.Client())
+func ModelProviderResolverFactory(name string, rawConfig json.RawMessage, handle framework.Handle) (framework.BBRPlugin, error) {
+	utilruntime.Must(inferencev1alpha1.AddToScheme(handle.Client().Scheme()))
+	utilruntime.Must(gatewayapiv1.Install(handle.Client().Scheme()))
```
nirrozenbaum: this was in the NewProviderResolver function.. should get back there.

cmd/main.go
```diff
@@ -37,7 +38,10 @@ func main() {
 
 	if err := runner.NewRunner().
 		WithExecutableName("ai-gateway-payload-processing").
-		// WithCustomCollectors(...). // THIS should be used for custom metrics exposed by our plugins
+		WithCustomControllers(
+			controllers.ProviderController(),
+			controllers.ModelController("", ""),
+		).
```
nirrozenbaum, line 47: can you move here also the legacy controller (take it out of the resolver)

cmd/main.go
```diff
@@ -37,7 +38,13 @@ func main() {
 
 	if err := runner.NewRunner().
 		WithExecutableName("ai-gateway-payload-processing").
-		// WithCustomCollectors(...). // THIS should be used for custom metrics exposed by our plugins
+		WithCustomControllers(
```
nirrozenbaum, line 41: can you move here also the legacy controller (take it out of the resolver)

test/e2e/e2e_suite_test.go
```diff
@@ -119,16 +89,14 @@ spec:
 		createProviderResources(p)
 	}
 
-	ginkgo.By("Waiting for plugin reconcilers to sync")
-	time.Sleep(10 * time.Second)
+	ginkgo.By("Waiting for controllers to create networking resources and routes to propagate")
+	time.Sleep(20 * time.Second)
```
nirrozenbaum, line 93: can we switch to using eventually instead of time.sleep (which is fragile)

nirrozenbaum (approved):
@yossiovadia left minor comments. 
don't want to block on them but would be good if we can address in a follow up

/lgtm
/approve

PR #322, opened 2026-06-11
added provider names

nirrozenbaum:
## Description
added provider names

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added support for Azure, Vertex, and Bedrock model providers and an OAuth2 auth option (alongside Simple and SigV4).
* **Refactor**
  * API format handling converted to a typed format for more reliable request/response detection and provider resolution.
* **Documentation/Schema**
  * CRD validation updated to explicitly allow oauth2 for auth type.
* **Tests**
  * Updated tests to use the new typed API format values.

fix #321

### DIFF
--- a/pkg/plugins/common/auth/auth.go
+++ b/pkg/plugins/common/auth/auth.go
@@ -1,0 +1,25 @@
+/*
+Copyright 2026 The opendatahub.io Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+type Auth string
+
+const (
+	Simple Auth = "simple"
+	SigV4  Auth = "sigv4"  // not used yet
+	OAuth2 Auth = "oauth2" // not used yet
+)
--- a/pkg/plugins/model-provider-resolver/external_model_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_model_reconciler.go
@@ -29,6 +29,7 @@ import (
 	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/observability/logging"
 
 	inferencev1alpha1 "github.com/opendatahub-io/ai-gateway-payload-processing/api/inference/v1alpha1"
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/apiformat"
 )
 
 const providerRequeueDelay = 5 * time.Second
@@ -103,7 +104,7 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1al
 	return &resolvedProviderRef{
 		provider:        providerInfo.provider,
 		targetModel:     ref.TargetModel,
-		apiFormat:       ref.APIFormat,
+		apiFormat:       apiformat.APIFormat(ref.APIFormat),
 		secretName:      secretName,
 		secretNamespace: secretNamespace,
 		config:          config,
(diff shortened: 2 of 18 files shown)


pkg/plugins/common/auth/auth.go
```diff
@@ -0,0 +1,25 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+type Auth string
```
noyitz, line 19: The `auth` package name collides with the existing `pkg/plugins/apikey-injection/auth` package. Having two `auth` packages at different import paths will require aliases everywhere they are used together. Consider naming this `authmode` or `authtype` to avoid the confusion.

noyitz:
Since we are formalizing the supported provider names, API formats, and auth types as typed constants, would be good to update `docs/providers/README.md` to reference these values. We already get questions from other teams about what is supported, and having the canonical list in one place (with the exact string values that go in the CRD) would save back and forth. The current docs still reference the old provider names (`bedrock-openai`, `azure-openai`, `vertex-openai`) and do not document the new ones or the `apiFormat` / `auth.type` values.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -103,7 +104,7 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref inferencev1al
 	return &resolvedProviderRef{
 		provider:        providerInfo.provider,
 		targetModel:     ref.TargetModel,
-		apiFormat:       ref.APIFormat,
+		apiFormat:       apiformat.APIFormat(ref.APIFormat),
```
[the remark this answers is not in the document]
nirrozenbaum: will add CRD validation in follow up

pkg/plugins/common/auth/auth.go
nirrozenbaum: updated the package of apikey-injection to `authgenerator`.

nirrozenbaum:
> Since we are formalizing the supported provider names, API formats, and auth types as typed constants, would be good to update `docs/providers/README.md` to reference these values. We already get questions from other teams about what is supported, and having the canonical list in one place (with the exact string values that go in the CRD) would save back and forth. The current docs still reference the old provider names (`bedrock-openai`, `azure-openai`, `vertex-openai`) and do not document the new ones or the `apiFormat` / `auth.type` values.

@noyitz yes, overall I agree. but that was not the intention of this PR. 
current PR was pushed to prepare the ground for working on auth mapping from auth type to the right api-key injection and completely detach it from provider. of course we need clear documentation about the allowed values, but atm these values are not really working yet (as long as the code is not updated first, the docs would just be wrong). 

this should be the other way around - update the code first. make sure it works with the auth types. then document it. 
we cannot document something that doesn't exist or doesn't work yet (and anyway it was not the purpose of this PR).

I suggest we defer the doc to a follow up PR.

yossiovadia (commented):
The apikey-injection plugin still maps auth generators by provider name (`provider.OpenAI`, `provider.Anthropic`, etc.), not by `auth.type`. The new `auth.Auth` type is defined in `common/auth` but not used in the plugin's generator selection — the mapping is still `provider → header`.

Is this intentional as incremental progress, or should this PR also switch the generator lookup to use `(provider, auth.type)`? If incremental, might be worth a TODO on the map in `plugin.go:73` so it doesn't get lost.

nirrozenbaum:
> Is this intentional as incremental progress, or should this PR also switch the generator lookup to use `(provider, auth.type)`? If incremental, might be worth a TODO on the map in `plugin.go:73` so it doesn't get lost.

@yossiovadia yes, that was intentional. also if we go to the direction of adding config for headerName and valuePrefix in auth config (on the CR), that means we won’t need to map based on provider. ideally we would like to completely remove provider from any mapping so it’s possible to add any provider 🙏🏼

we have issue #317 to track it, so it doesn’t get lost

yossiovadia (approved):
/lgtm

PR #323, opened 2026-06-12
feat: add spec.modelName to decouple client name from metadata.name

yossiovadia:
## Summary

Implements #314. Adds optional `spec.modelName` field to ExternalModel CRD so the client-facing model name can differ from `metadata.name`.

**Problem:** Kubernetes `metadata.name` has character restrictions (lowercase, no dots/colons/slashes, RFC 1123) that prevent using real model identifiers like `anthropic.claude-3-opus` or `gpt-4o:2025-01-01` as CR names.

**Solution:** When `spec.modelName` is set, it becomes the client-facing name. When omitted, `metadata.name` is used (backward compatible).

```yaml
apiVersion: inference.opendatahub.io/v1alpha1
kind: ExternalModel
metadata:
  name: gpt4              # K8s identifier
spec:
  modelName: gpt-4o       # client-facing name in request body
  externalProviderRefs:
    - ref: { name: openai }
      targetModel: gpt-4o
      apiFormat: openai-chat
```

## Changes

- **CRD**: new optional `modelName` field on `ExternalModelSpec` (max 253 chars)
- **Store**: `externalModelInfo` tracks `modelName`
- **Reconciler**: populates `modelName` from `spec.modelName` or falls back to `metadata.name`
- **ProcessRequest**: validates body `model` against `modelName` before weight selection
- **Tests**: `TestModelReconciler_ModelNameOverride` + happy path assertion
- **Samples**: OpenAI sample updated to show `modelName`

## Size note

~60 lines of production code. Rest is generated files (deepcopy + CRD YAML) and tests.

## Test plan

- [x] `make test-unit` — all tests pass (78% coverage)
- [x] `make lint` — 0 issues
- [x] `make generate manifests` — generated files up to date
- [x] `TestModelReconciler_ModelNameOverride` — verifies spec.modelName overrides metadata.name
- [x] `TestModelReconciler_HappyPath` — verifies default to metadata.name

Closes #314

PR #313, opened 2026-06-10
fix: address #308 review — model name check bug, pointer slices, style

yossiovadia:
Follow-up to #308 per Nir review comments (PR merged before review complete).

Bug fix:
- Model name validation now compares against ExternalModel CR name (client-facing name), not targetModel
- Validation moved BEFORE weighted provider selection

Style:
- Pointer slices, pass ref as pointer, descriptive names, k8s.io/utils/ptr
- Restore removed test assertions, add TestProcessRequest_ModelMismatch

Test plan:
- make test-unit passes (76.6% coverage)
- make lint 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests are now validated against the resolved external model name before provider selection; mismatches return a clear not-found error.
  * Unavailable provider references are skipped until available, improving reliability of provider resolution.

* **Chores**
  * Internal reference handling and in-memory storage updated for more consistent provider resolution and selection behavior.

* **Tests**
  * Tests strengthened for secret/namespace handling, selection distribution, and model-mismatch scenarios.

PR #326, opened 2026-06-12
fix: pointer slices, descriptive names, model mismatch test (#308 follow-up)

yossiovadia:
Supersedes #313. Rebased cleanly on main after #322 and #323 merged.

Per Nir's review on #308:
- Use `[]*resolvedProviderRef` (pointer slice) to avoid copying structs
- Pass ref as pointer to `resolveRef`
- Use descriptive variable name (`resolvedRef` instead of `rr`)
- Use `k8s.io/utils/ptr.To` instead of custom `intPtr`
- Add `TestProcessRequest_ModelMismatch`

All tests pass, lint clean.

## Release Notes

* **Bug Fixes**
  * Added validation to ensure request model names match configured external model entries.

* **Refactor**
  * Optimized internal handling of external model provider references.


nirrozenbaum (approved):
/lgtm

PR #329, opened 2026-06-15
update api key injection plugin mapping by auth type

nirrozenbaum:
## Description
update the apijey injection mapping to be by auth type (from the CR).
Fix: #317 

## Merge criteria:

## Release Notes

* **New Features**
  * Added authentication-type-based API key injection with support for Simple, SigV4, and OAuth2
  * Added authentication metadata propagation through model/provider resolution
  * Added support for Azure OpenAI, Vertex OpenAI, and Bedrock OpenAI providers

* **Refactor**
  * Updated authentication header generation to be configuration/credentials-driven with sensible defaults

* **Tests**
  * Expanded unit coverage for auth header generation, extraction, and auth defaulting behavior

### DIFF
--- a/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
+++ b/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
@@ -19,38 +19,78 @@ package authgenerator
 import (
 	"fmt"
 
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/auth"
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/state"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/bbr/framework"
 )
 
 // apiKeyField is the field name in the credentials map that holds the API key.
-const apiKeyField = "api-key"
+const (
+	apiKeyField            = "api-key"
+	defaultAuthHeader      = "Authorization"
+	defaultAuthValuePrefix = "Bearer "
+)
 
 // compile-time interface check
 var _ AuthHeadersGenerator = &SimpleAuthGenerator{}
 
+func NewSimpleAuthGenerator() *SimpleAuthGenerator {
+	return &SimpleAuthGenerator{}
+}
+
 // SimpleAuthGenerator generates a single auth header from an API key.
-// HeaderName is the HTTP header (e.g. "Authorization", "x-api-key").
-// HeaderValuePrefix is prepended to the key (e.g. "Bearer "); use "" for raw keys.
+// ExtractRequestData resolves the header name and value prefix from the model
+// config in CycleState (defaults are injected by the provider reconciler).
+// GenerateAuthHeaders reads those values from the merged credentials map.
 // Requires the credentials map to contain an "api-key" field.
-type SimpleAuthGenerator struct {
-	HeaderName        string
-	HeaderValuePrefix string
+type SimpleAuthGenerator struct{}
+
+// ExtractRequestData resolves the auth header name and value prefix from the
+// model config in CycleState. Provider-specific defaults (e.g. "x-api-key" for
+// Anthropic) are injected into the config by the provider reconciler. If not
+// set, falls back to "Authorization" with "Bearer " prefix.
+func (g *SimpleAuthGenerator) ExtractRequestData(cycleState *framework.CycleState, _ *framework.InferenceRequest) (map[string]string, error) {
+	config, err := framework.ReadCycleStateKey[map[string]string](cycleState, state.ModelConfigKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract config from cycle state - %w", err)
+	}
+
+	authHeader := defaultAuthHeader
+	authValuePrefix := defaultAuthValuePrefix
+
+	if headerName, ok := config[auth.SimpleAuthHeaderName]; ok {
+		authHeader = headerName
+		authValuePrefix = "" // non-default header implies no prefix unless explicitly set
+	}
+	if valuePrefix, ok := config[auth.SimpleAuthValuePrefix]; ok {
+		authValuePrefix = valuePrefix
+	}
+
+	return map[string]string{
+		auth.SimpleAuthHeaderName:  authHeader,
+		auth.SimpleAuthValuePrefix: authValuePrefix,
+	}, nil
 }
 
-// ExtractRequestData is a no-op for SimpleAuthGenerator — API key auth doesn't need request data.
-func (g *SimpleAuthGenerator) ExtractRequestData(_ *framework.CycleState, _ *framework.InferenceRequest) (map[string]string, error) {
-	return nil, nil
-}
-
-// GenerateAuthHeaders extracts the "api-key" field from credentials and returns
+// GenerateAuthHeaders extracts the relevant fields from credentialsData and returns
 // the header name and formatted value. Returns an error if the field is missing.
-func (g *SimpleAuthGenerator) GenerateAuthHeaders(credentials map[string]string) (map[string]string, error) {
-	apiKey, ok := credentials[apiKeyField]
+func (g *SimpleAuthGenerator) GenerateAuthHeaders(credentialsData map[string]string) (map[string]string, error) {
+	apiKey, ok := credentialsData[apiKeyField]
 	if !ok {
 		return nil, fmt.Errorf("credentials missing required field %s", apiKeyField)
 	}
 
+	headerName, ok := credentialsData[auth.SimpleAuthHeaderName]
+	if !ok {
+		return nil, fmt.Errorf("credentials missing required field %s", auth.SimpleAuthHeaderName)
+	}
+
+	valuePrefix, ok := credentialsData[auth.SimpleAuthValuePrefix]
+	if !ok {
+		return nil, fmt.Errorf("credentials missing required field %s", auth.SimpleAuthValuePrefix)
+	}
+
 	return map[string]string{
-		g.HeaderName: fmt.Sprintf("%s%s", g.HeaderValuePrefix, apiKey),
+		headerName: fmt.Sprintf("%s%s", valuePrefix, apiKey),
 	}, nil
 }
--- a/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
+++ b/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
@@ -60,10 +60,10 @@ func (g *SimpleAuthGenerator) ExtractRequestData(cycleState *framework.CycleStat
 
 	if headerName, ok := config[auth.SimpleAuthHeaderName]; ok {
 		authHeader = headerName
-		authValuePrefix = "" // non-default header implies no prefix unless explicitly set
-	}
-	if valuePrefix, ok := config[auth.SimpleAuthValuePrefix]; ok {
-		authValuePrefix = valuePrefix
+		authValuePrefix = ""                                           // non-default header implies no prefix unless explicitly set
+		if valuePrefix, ok := config[auth.SimpleAuthValuePrefix]; ok { // valuePrefix can be set only if header name is set
+			authValuePrefix = valuePrefix
+		}
 	}
 
 	return map[string]string{
--- a/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
+++ b/pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
@@ -58,7 +58,7 @@ func (g *SimpleAuthGenerator) ExtractRequestData(cycleState *framework.CycleStat
 	authHeader := defaultAuthHeader
 	authValuePrefix := defaultAuthValuePrefix
 
-	if headerName, ok := config[auth.SimpleAuthHeaderName]; ok {
+	if headerName, ok := config[auth.SimpleAuthHeaderName]; ok && headerName != "" {
 		authHeader = headerName
 		authValuePrefix = ""                                           // non-default header implies no prefix unless explicitly set
 		if valuePrefix, ok := config[auth.SimpleAuthValuePrefix]; ok { // valuePrefix can be set only if header name is set
--- a/pkg/plugins/model-provider-resolver/external_provider_reconciler.go
+++ b/pkg/plugins/model-provider-resolver/external_provider_reconciler.go
@@ -19,6 +19,7 @@ package model_provider_resolver
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -30,6 +31,16 @@ import (
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/auth"
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/provider"
 )
+
+// defaultAuthHeaders maps providers that use a non-standard auth header name.
+// Providers not listed here fall back to "Authorization" with "Bearer " prefix
+// in the auth generator.
+var defaultAuthHeaders = map[string]string{
+	provider.Anthropic: "x-api-key",
+	provider.Azure:     "api-key",
+	// TODO to be removed
+	provider.AzureOpenAI: "api-key",
+}
 
 type externalProviderReconciler struct {
 	client.Reader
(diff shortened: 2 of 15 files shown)


nirrozenbaum:
cc: @noyitz @yossiovadia @yehuditkerido @noalimoy 
please pay attention that this PR changes the mapping of apikey-injection. it's no longer based on provider, but rather on the CR-> auth.type. 
once we remove provider mapping also from api-translation, the provider can become a free text that has no meaning (just name to display to users, no coupling to anything).

@yehuditkerido this is going to affect also OAuth2 PR which should align with this behavior.

yossiovadia:
## OpinAI PR Review

**PR:** #329
**Title:** PR #329
**Author:** @nirrozenbaum
**Verdict:** COMMENT
**Risk:** MEDIUM-HIGH
**Base branch:** main

---

## Summary

This PR refactors the API key injection plugin to select authentication header generators by auth type (`simple`, `sigv4`) from the ExternalProvider/ExternalModel CRD's `auth.type` field, instead of by provider name (e.g., `openai`, `anthropic`). Provider-specific auth header defaults (like `x-api-key` for Anthropic, `api-key` for Azure) are now injected into the config map during provider reconciliation via `buildConfigWithDefaults`, and the `SimpleAuthGenerator` reads these from CycleState at request time instead of using hardcoded struct fields.

This is a clean decoupling step that moves auth configuration from code to CR-driven config, aligning with the project's direction of making provider names free-text. However, it introduces a tighter coupling between the reconciler's config injection and the plugin's runtime expectations, creating new failure modes that warrant careful analysis.

## Risk Assessment

**MEDIUM-HIGH** — This PR changes the behavioral contract between the reconciler and the apikey-injection plugin in two ways: (1) auth generator selection is now driven by `auth.type` rather than provider name, and (2) the `SimpleAuthGenerator` now requires `authHeaderName` and `authValuePrefix` keys in the config map read from CycleState. Any break in the chain (reconciler must inject defaults → config must be written to CycleState → generator reads config) causes auth failures at runtime. The risk is elevated by the lack of fail-safe behavior when defaults are missing and the potential for version skew during rolling upgrades.

---

## Previously Identified (by other reviewers)

| Finding | Disposition |
|---------|-------------|
| @coderabbitai: Map pre-allocation undersizing at `plugin.go` line 141 — `make(map[string]string, len(credentialsData))` should be `len(credentialsData)+len(extraData)` | Valid minor optimization. Not a correctness bug — `maps.Copy` handles growth. **Actionable:** Fix allocation size for correctness-of-intent. |
| @coderabbitai: Missing CR auth-type validation at reconciler level | Mitigated by CRD-level validation: `+kubebuilder:validation:Enum=simple;sigv4;oauth2` on `AuthConfig.Type` (`api/inference/v1alpha1/common_types.go` line 24). Invalid values are rejected at the Kubernetes API level. In-code validation remains a defense-in-depth concern. |

---

## New Findings

### Finding 1: `oauth2` CRD Enum Accepted but No Runtime Handler
**Severity:** Low (Known Limitation)
**Impact:** If a user creates an ExternalProvider with `auth.type: oauth2`, it passes CRD validation but fails at runtime with `unsupported authType - 'oauth2'` (HTTP 500).
**Details:** `auth.go` line 24 defines `OAuth2 Auth = "oauth2" // not used yet` and the CRD allows it via `Enum=simple;sigv4;oauth2`. The author is aware (comments reference alignment with the OAuth2 PR by @yehuditkerido).
**Recommendation:** Either (a) remove `oauth2` from the CRD enum until supported, preventing user confusion, or (b) return a descriptive error message: `"oauth2 auth type is defined but not yet implemented — see issue #XXX"` instead of the generic "unsupported authType" error.

### Finding 2: `SimpleAuthGenerator.ExtractRequestData` Introduces New Required CycleState Dependency
**Severity:** Medium
**Impact:** `ExtractRequestData` now reads `state.ModelConfigKey` from CycleState (`simple_auth_generator.go` lines 57-59) and returns an error if the key is not found. Previously, `ExtractRequestData` was effectively a no-op for `SimpleAuthGenerator` — it returned no data and no error. After this PR, a missing `ModelConfigKey` in CycleState causes `ExtractRequestData` to return an error, which propagates as an auth failure (HTTP 500) in the plugin's `PostScheduling` hook.
**Failure chain:** If the model-provider-resolver plugin does not write `ModelConfigKey` to CycleState (e.g., due to a code path change, a bug in scheduling, or version skew), all requests to external models using `simple` auth will fail with an opaque error.
**Mitigation present:** The model-provider-resolver always writes `ModelConfigKey` at line 185 (`cycleState.Write(state.ModelConfigKey, ref.config)`) for external models. The test `TestSimpleExtractRequestData_MissingModelConfigKey` covers the error path.
**Recommendation:** Add a defensive log message at WARN level when `ModelConfigKey` is absent, explicitly stating `"ModelConfigKey not found in CycleState — this indicates model-provider-resolver did not run or did not write config"`, to aid operational debugging.

### Finding 3: Rolling Upgrade / Version Skew Safety
**Severity:** Medium-High
**Impact:** During a rolling upgrade, the new apikey-injection plugin code may run against CycleState populated by an old model-provider-resolver that has not yet been upgraded. The old reconciler does not call `buildConfigWithDefaults`, so the config map in CycleState will lack `authHeaderName` and `authValuePrefix` keys.
**Analysis of failure behavior in `SimpleAuthGenerator`:**
- `simple_auth_generator.go` line 60: `authHeaderName := config[authHeaderNameKey]` — if key is missing, Go returns `""` (empty string).
- `simple_auth_generator.go` line 61: `authValuePrefix := config[authValuePrefixKey]` — same, returns `""`.
- In `GenerateAuthHeaders`, `authHeaderName` being empty means the auth credential would be set on an empty-string header key, which is semantically broken — the upstream LLM provider would not receive a valid auth header.
**Recommendation:** Add an explicit check in `ExtractRequestData` or `GenerateAuthHeaders`: if `authHeaderName` is empty after reading from config, return an error with message `"authHeaderName is missing from model config — ensure the reconciler has been upgraded to inject auth defaults via buildConfigWithDefaults"`. This converts a silent misconfiguration into a clear operational signal.

### Finding 4: Reconciler-Side `buildConfigWithDefaults` — Completeness and Fail-Safe Behavior
**Severity:** Medium
**Impact:** The `buildConfigWithDefaults` function in the reconciler is the single source of truth for provider-specific auth header defaults. If a new provider is added to the system without updating this function's switch/map, the config will lack `authHeaderName` and `authValuePrefix`, causing the failure described in Finding 3.
**Questions for the author:**
1. Is `buildConfigWithDefaults` tested independently with unit tests covering each provider's default injection?
2. What is the default/fallback behavior when a provider name is not recognized by `buildConfigWithDefaults`? Does it inject a sensible default (e.g., `Authorization` header with `Bearer` prefix) or produce an empty map?
3. Is there a test that verifies adding a new provider without updating `buildConfigWithDefaults` produces a clear error rather than silent auth failure?
**Recommendation:** Add a fallback in `buildConfigWithDefaults` that injects `authHeaderName: "Authorization"` and `authValuePrefix: "Bearer "` for unrecognized providers, matching the most common LLM API auth convention. Log a WARN when falling back. Add a unit test for the fallback path.

### Finding 5: Credential Map Dual-Use — Secret Data Mixed with Runtime Config
**Severity:** Low-Medium
**Impact:** After the merge at `plugin.go` lines 141-144, the `credentialsData` map contains both sensitive Secret data (e.g., `api-key: "sk-..."`) and non-sensitive runtime config (`authHeaderName: "x-api-key"`, `authValuePrefix: ""`). This merged map is passed to `GenerateAuthHeaders`.
**Concerns:**
- **Logging/leakage:** If any downstream code logs or serializes this map (e.g., in error messages, debug logs, or metrics), it would leak API keys alongside config values. Verified that `GenerateAuthHeaders` does not log the map, but this is a fragile guarantee.
- **Key collision:** If a user creates a Kubernetes Secret with data keys `authHeaderName` or `authValuePrefix`, the runtime values from `extraData` overwrite them (due to `maps.Copy(merged, extraData)` running second). This is the correct precedence (runtime config wins), but could be confusing. The camelCase naming convention makes collision unlikely but not impossible.
**Recommendation:** (a) Document in code comments that the merged map contains sensitive data and must not be logged. (b) Consider prefixing runtime config keys with a namespace like `__auth.` to eliminate collision risk (e.g., `__auth.headerName`).

### Finding 6: Existing Gotcha — Secret Key Validation Not Addressed
**Severity:** Low (Pre-existing)
**Impact:** The existing concern (`gotcha_externalprovider_secret_key_not_validated`) about whether the referenced Secret actually contains the expected data key (e.g., `api-key`) is not addressed or worsened by this PR. The `GenerateAuthHeaders` in `SimpleAuthGenerator` reads `credData[credentialKey]` where `credentialKey` defaults to `"api-key"` — if the Secret doesn't have this key, an empty credential value is silently used, resulting in a request to the upstream provider with an empty auth header value.
**Recommendation:** This is pre-existing and out of scope for this PR, but given that this PR refactors the credential handling path, it would be a good opportunity to add a validation check: if the credential value is empty after reading from the merged map, return an error rather than sending an unauthenticated request upstream.

---

## Downstream Plugin Impact Analysis

| Plugin | CycleState Keys Used | Impact from This PR | Risk |
|--------|---------------------|---------------------|------|
| **model-provider-resolver** | Writes: `ModelConfigKey`, `AuthTypeKey` | **Writer** — must now populate config with `authHeaderName`/`authValuePrefix` via `buildConfigWithDefaults` during reconciliation. New responsibility. | Medium — if defaults are missing, downstream auth fails silently |
| **apikey-injection** | Reads: `AuthTypeKey`, `ModelConfigKey`, Secret data | **Modified** — generator selection changed from provider-name to auth-type. `SimpleAuthGenerator.ExtractRequestData` now reads config from CycleState. | Medium — new failure mode if `ModelConfigKey` absent |
| **Other scheduling plugins** | May read `ModelConfigKey` | **No direct impact** — `ModelConfigKey` data shape is a `map[string]string` and the added keys are additive. Plugins that iterate over all config keys may encounter the new auth keys but should be unaffected unless they have key-name assumptions. | Low |
| **sigv4-injection** (if separate) | Reads: `AuthTypeKey` | **No impact** — `AuthTypeKey` value `sigv4` routes to existing SigV4 generator. | None |

---

## Merge Order in `plugin.go` — Confirmed Correct

The merge at lines 142-143:
```go
maps.Copy(merged, credentialsData)  // first: Secret data
maps.Copy(merged, extraData)        // second: runtime config overwrites
```
Runtime-computed auth config (from CycleState) takes precedence over Secret data. This is the correct and safe ordering — runtime config should not be overridable by user-controlled Secret data, as that could allow a Secret to redirect auth headers to attacker-controlled endpoints.

---

## Recommendations Summary

| # | Finding | Severity | Action |
|---|---------|----------|--------|
| 1 | `oauth2` enum with no handler | Low | Remove from CRD enum or improve error message |
| 2 | New `ModelConfigKey` dependency | Medium | Add WARN-level log with diagnostic message |
| 3 | Rolling upgrade version skew | Medium-High | **Add empty-string guard on `authHeaderName`** — fail explicitly rather than silently |
| 4 | `buildConfigWithDefaults` completeness | Medium | Add fallback defaults for unrecognized providers; add unit tests |
| 5 | Credential map dual-use | Low-Medium | Add code comments prohibiting logging; consider key namespacing |
| 6 | Secret key validation (pre-existing) | Low | Out of scope, but opportunistic fix recommended |

---

===PR_VERDICT===
VERDICT: COMMENT
RISK: MEDIUM-HIGH

pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
```diff
@@ -19,38 +19,78 @@ package authgenerator
...
+		return nil, fmt.Errorf("failed to extract config from cycle state - %w", err)
+	}
+
+	authHeader := defaultAuthHeader
+	authValuePrefix := defaultAuthValuePrefix
+
+	if headerName, ok := config[auth.SimpleAuthHeaderName]; ok {
+		authHeader = headerName
+		authValuePrefix = ""                                           // non-default header implies no prefix unless explicitly set
+		if valuePrefix, ok := config[auth.SimpleAuthValuePrefix]; ok { // valuePrefix can be set only if header name is set
```
noyitz, line 64: If `ModelConfigKey` is missing from CycleState this returns an error and fails the entire request. Could this legitimately be absent in edge cases (legacy ExternalModel without ExternalProvider, or test setups that only write provider + creds)? If so, falling back to defaults (Authorization / Bearer) instead of erroring would be safer and consistent with the zero-config philosophy for known providers.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -27,8 +28,20 @@ import (
...
 	inferencev1alpha1 "github.com/opendatahub-io/ai-gateway-payload-processing/api/inference/v1alpha1"
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/auth"
+	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/provider"
 )
 
+// defaultAuthHeaders maps providers that use a non-standard auth header name.
+// Providers not listed here fall back to "Authorization" with "Bearer " prefix
+// in the auth generator.
+var defaultAuthHeaders = map[string]string{
+	provider.Anthropic: "x-api-key",
```
noyitz, line 39: If the MaaS controller needs to know the default auth header per provider (e.g., to generate the correct Authorino identity source for inbound auth, as discussed in RHOAIENG-67493), it would not be able to reuse this map since it lives in the provider reconciler. Consider moving `defaultAuthHeaders` to the `auth` or `provider` package so it can be shared across components.

pkg/plugins/apikey-injection/auth-generator/simple_auth_generator.go
nirrozenbaum: > Could this legitimately be absent in edge cases (legacy ExternalModel without ExternalProvider, or test setups that only write provider + creds)

legacy ExternalModel is not reconciled and handled by model resolver, it's reconciled by a dedicated migration controller that creates the new CRs which in turn are handled correctly.
having ExternalProvider alone is not enough to serve a model (it's just the provider).

> falling back to defaults (Authorization / Bearer) instead of erroring would be safer and consistent with the zero-config philosophy for known providers.

we do have a default (Authorization/Bearer) in the simple auth generator in case the extra fields were not set.. see defaults in SimpleAuthGenerator.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
nirrozenbaum: I don't mind moving it, but.. I don't see a reason for Authorino to reference IPP only for this small mapping.
go principles say - [A little copying is better than a little dependency.](https://go-proverbs.github.io/)

nirrozenbaum:
@yossiovadia

> finding (1) - oauth2 CRD Enum Accepted but No Runtime Handler

yes, this is known. oauth2 PR is currently WIP, but since IPP is not released yet I think it's a valid option to keep the enum value rather than back and forth.. (remove it, then readd it). shouldn't be a big deal for now since this is a new field and if anyone by mistake uses oauth2 in the meantime they will get an error. this is the next PR in the pipeline.

> finding (2) - SimpleAuthGenerator.ExtractRequestData Introduces New Required CycleState Dependency:

yes, this is the whole point of the CR :). we want to switch into mapping auth to the selection auth generator. resolver plugin is the one that always runs first and resolves the model/provider related fields and then adds them into CycleState. that was also the case when the mapping was based on provider so nothing changes from that perspective.

> Finding 3 - Rolling Upgrade / Version Skew Safety
Severity: Medium-High
Impact: During a rolling upgrade, the new apikey-injection plugin code may run against CycleState populated by an old model-provider-resolver that has not yet been upgraded. 

This looks like an AI slop.

> Finding 4 - Reconciler-Side buildConfigWithDefaults — Completeness and Fail-Safe Behavior

another AI slop. there are defaults in the simple auth generator.

> finding 5 - Credential Map Dual-Use — Secret Data Mixed with Runtime Config

the comment itself is correct, but IMO shouldn't be flagged. we pass to auth generator the fields it needs to generate auth headers. whether it's confidential api key or just the header name. both are needed to generate the auth headers. 
the concern is for possible leakage in logging, but the PR is not logging none of those fields. I'm not concerned about this point.

> finding 6 -  Existing Gotcha — Secret Key Validation Not Addressed

not relevant to this PR. we can open a dedicated issue to track it.

noyitz (approved):
lgtm

PR #331, opened 2026-06-16
feat: migrate framework imports to llm-d IPP

noyitz:
## Summary

- Migrate all plugin and runner imports from `sigs.k8s.io/gateway-api-inference-extension` to `github.com/llm-d/llm-d-inference-payload-processor`
- This switches the MaaS binary to use the IPP runner, which includes the streaming response body mode framework
- 25 files updated, no logic changes — pure import migration

**Depends on:** llm-d/llm-d-inference-payload-processor#169 (response body mode framework)

## Why

The MaaS binary was using the old BBR runner from `gateway-api-inference-extension`, which always buffers all response body chunks. The active framework at `llm-d/llm-d-inference-payload-processor` has streaming support (per-profile conditional buffering, ChunkProcessor) that only takes effect when the ODH plugins use the IPP types.

## Import mapping

| Old | New |
|-----|-----|
| `gateway-api-inference-extension/pkg/bbr/framework` | `llm-d-inference-payload-processor/pkg/framework/interface/requesthandling` |
| `gateway-api-inference-extension/pkg/epp/framework/interface/plugin` | `llm-d-inference-payload-processor/pkg/framework/interface/plugin` |
| `gateway-api-inference-extension/pkg/common/error` | `llm-d-inference-payload-processor/pkg/common/error` |
| `gateway-api-inference-extension/pkg/common/observability/logging` | `llm-d-inference-payload-processor/pkg/common/observability/logging` |
| `gateway-api-inference-extension/cmd/bbr/runner` | `llm-d-inference-payload-processor/cmd/runner` |

## go.mod

- Added `github.com/llm-d/llm-d-inference-payload-processor` (local replace directive until version is published)
- `gateway-api-inference-extension` is fully removed from go.mod

## Test plan

- [x] `go build ./...` passes
- [x] All plugin tests pass (`api-translation`, `apikey-injection`, `model-provider-resolver`, `nemo`)
- [x] `legacymigration` test requires envtest (kubebuilder binaries) — pre-existing, unrelated to this change

## Summary by CodeRabbit

* **Refactor**
  * Migrated plugin wiring and request/response processing interfaces from Gateway API Inference Extension to LLM-D Inference Payload Processor across API translation, API key injection, model provider resolution, and NeMo request/response guards (including cycle-state handling changes).
  * Updated translators and auth/header generation to use the payload-processor shared error types (OpenAI/Anthropic/Vertex/Bedrock).

* **Chores**
  * Updated Helm and e2e configurations from `upstreamBbr` to `upstreamIpp`, including RBAC naming and payload-processing chart wiring; refreshed build metadata and Go module dependencies; updated `.gitignore`.

* **Tests**
  * Updated unit and e2e test scaffolding to use the new request/response and cycle-state types.


nirrozenbaum (commented):
**go.mod: local `replace` directive must be removed before merge**

The `replace` directive points to a local filesystem path:
```
replace github.com/llm-d/llm-d-inference-payload-processor => ../llm-d-inference-payload-processor
```

This will break for anyone cloning the repo. The module is already published — remove the `replace` directive and use a real version in the `require` block instead:

```go
require (
    github.com/llm-d/llm-d-inference-payload-processor v0.0.0-20260609111331-11891e132621
)
```

To fix:
1. Delete the `replace github.com/llm-d/llm-d-inference-payload-processor => ../llm-d-inference-payload-processor` line from `go.mod`
2. Run `go get github.com/llm-d/llm-d-inference-payload-processor@latest`
3. Run `go mod tidy`
4. Verify with `go build ./...`

noyitz:
Fixed — replaced local path with fork URL (`github.com/noyitz/llm-d-inference-payload-processor@ca1985b`).

Can't point to upstream main yet because the `ResponseBodyMode` types are in PR [llm-d/llm-d-inference-payload-processor#169](https://github.com/llm-d/llm-d-inference-payload-processor/pull/169) which hasn't merged. The `replace` directive will be removed once #169 lands upstream.

nirrozenbaum:
**Plugin configuration: `--plugin` CLI args no longer supported in llm-d-inference-payload-processor**

The upstream runner in `llm-d-inference-payload-processor` no longer accepts `--plugin` CLI flags. Plugins are now configured via a YAML config file, loaded with `--config-file <path>` (or `--config-text` for inline).

The expected config format is:

```yaml
apiVersion: llm-d.ai/v1alpha1
kind: PayloadProcessorConfig
plugins:
  - type: body-field-to-header
    name: model-extractor
    parameters:
      fieldName: model
      headerName: X-Gateway-Model-Name
  - type: model-provider-resolver
  - type: api-translation
  - type: apikey-injection

profiles:
  default:
    request:
      plugins:
        - pluginRef: model-extractor
        - pluginRef: model-provider-resolver
        - pluginRef: api-translation
        - pluginRef: apikey-injection
    response:
      plugins:
        - pluginRef: api-translation
```

**What needs to change in this PR:**

1. Create a `ConfigMap` template in the Helm chart (`deploy/payload-processing/templates/`) that renders this YAML config from `values.yaml`:
   ```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: {{ .Values.upstreamIpp.ipp.name }}-plugin-config
   data:
     config.yaml: |
       # rendered plugin config from values
   ```

2. Mount the ConfigMap into the Deployment as a volume:
   ```yaml
   volumes:
     - name: plugin-config
       configMap:
         name: {{ .Values.upstreamIpp.ipp.name }}-plugin-config
   containers:
     - volumeMounts:
         - name: plugin-config
           mountPath: /etc/payload-processor
           readOnly: true
   ```

3. Pass the config file path to the runner via `--config-file /etc/payload-processor/config.yaml` instead of `--plugin` flags.

4. Adapt the `values.yaml` plugin list format — note the field is `parameters` (not `json`), and the processing order is defined in the `profiles` section.

noyitz:
Thanks — you're right about the config format change. Here's where we stand:

**Fixed so far:**
- ✅ `go.mod`: using published upstream version `v0.0.0-20260609111331-11891e132621`, no replace directive
- ✅ `Dockerfile`: updated to use `github.com/llm-d/llm-d-inference-payload-processor/version` for ldflags, renamed binary from `/bbr` to `/payload-processor`

**Still needed (Helm chart migration):**
The chart currently depends on the upstream `body-based-routing` chart from `gateway-api-inference-extension` (v1.5.0-rc.1) which renders the Deployment with `--plugin` CLI flags. The IPP runner expects `--config-file` with a YAML config instead.

Options:
1. **Replace the chart dependency** — point to the IPP's `payload-processor` chart (`config/charts/payload-processor`) which already has the ConfigMap + `--config-file` pattern
2. **Inline the templates** — remove the upstream dependency and add our own Deployment + ConfigMap templates directly

Which approach do you prefer? Option 1 is cleaner but means we depend on the IPP chart being published. Option 2 gives us full control.

nirrozenbaum:
option 1 please - replace the chart dependency to llm-d-inference-payload-processor

deploy/payload-processing/values.yaml
```diff
@@ -15,28 +17,57 @@ upstreamBbr:
...
+    # Format follows PayloadProcessorConfig (llm-d.ai/v1alpha1).
+    customConfig:
+      plugins:
+        - type: body-field-to-header
+          name: model-extractor
+          parameters:
+            fieldName: model
+            headerName: X-Gateway-Model-Name
+        - type: model-provider-resolver
+          name: model-provider-resolver
```
nirrozenbaum: remove plugin names when the name is identical to type

deploy/payload-processing/values.yaml
```diff
@@ -15,28 +17,57 @@ upstreamBbr:
...
+      requestBody: true
+      requestTrailers: true
+      responseHeaders: true
+      responseBody: true
+      responseTrailers: true
+
     istio:
       envoyFilter:
         operation: INSERT_AFTER
+        anchorSubFilter: ""
```
nirrozenbaum: remove

deploy/payload-processing/values.yaml
```diff
@@ -15,28 +17,57 @@ upstreamBbr:
...
+          plugins:
+            request:
+              - pluginRef: model-extractor
+              - pluginRef: model-provider-resolver
+              - pluginRef: api-translation
+              - pluginRef: apikey-injection
+            response:
+              - pluginRef: api-translation
+
+    tracing:
```
nirrozenbaum: remove tracing section

deploy/payload-processing/values.yaml
```diff
@@ -15,28 +17,57 @@ upstreamBbr:
...
+      otelServiceName: "inference.opendatahub.io/payload-processor"
+      otelExporterEndpoint: "http://localhost:4317"
+      sampling:
+        sampler: "parentbased_traceidratio"
+        samplerArg: "0.1"
 
   provider:
     name: istio
+
+    supportedEvents:
```
nirrozenbaum: remove section. 

more generally remove sections that are 100% duplication of ipp (no need to repeat)

pkg/plugins/model-provider-resolver/plugin_test.go
```diff
@@ -53,35 +54,35 @@ func TestProcessRequest_ModelResolved(t *testing.T) {
...
+	cs := pluginpkg.NewCycleState()
+	req := requesthandling.NewInferenceRequest()
 	req.Headers[":path"] = "/" + extNS + "/" + extName + "/v1/chat/completions"
 	req.Body["model"] = extName
 
 	err := plugin.ProcessRequest(context.Background(), cs, req)
 	require.NoError(t, err)
 
-	actualModel, err := framework.ReadCycleStateKey[string](cs, state.ModelKey)
+	actualModel, err := pluginpkg.ReadCycleStateKey[string](cs, state.ModelKey)
```
nirrozenbaum: nit: keep imports consistent. for this you need to change to variable name from plugin to instance or something like it.

deploy/payload-processing/values.yaml
nirrozenbaum: do the same also in the e2e customConfig

deploy/payload-processing/values.yaml
```diff
@@ -1,42 +1,34 @@
...
-      # Ordering: apikey-injection must be last. SigV4 signs request.Body —
-      # do not add body-mutating plugins after this point.
-      - type: apikey-injection
-        name: apikey-injection
-
-  provider:
-    name: istio
-    istio:
-      envoyFilter:
-        operation: INSERT_AFTER
```
nirrozenbaum, line 39: need to bring this back (in llm-d provider = none)

deploy/payload-processing/values.yaml
```diff
@@ -1,42 +1,34 @@
...
-
-  inferenceGateway:
-    name: maas-default-gateway
```
nirrozenbaum, line 42: need to this back as well

deploy/payload-processing/Chart.yaml
```diff
@@ -1,16 +1,16 @@
...
 
 dependencies:
-  - name: body-based-routing
-    repository: oci://registry.k8s.io/gateway-api-inference-extension/charts
-    version: v1.5.0-rc.1
-    # Upstream chart; overrides go under upstreamBbr in this chart’s values.yaml.
-    alias: upstreamBbr
+  - name: payload-processor
+    # Local path to the IPP chart. Update to an OCI/repo URL once the chart is published.
+    repository: file://../../../llm-d-inference-payload-processor/config/charts/payload-processor
```
nirrozenbaum: this is the only concerning point in current PR. 
need to push for a pre-release including image of the chart that can be referenced, which should unblock this PR.

api/inference/v1alpha1/externalmodel_types.go
nirrozenbaum, line 47:
```suggestion
	// ModelName is the client-facing model name used in inference request bodies.
	// Clients send this value in the "model" field of inference requests.
	// Defaults to metadata.name if not set. Use this field when the desired
	// model name contains characters not allowed in Kubernetes resource names
	// (dots, colons, slashes, uppercase).
	// +optional
	// +kubebuilder:validation:MaxLength=253
```

test/e2e/scripts/e2e-values.yaml
```diff
@@ -1,22 +1,33 @@
...
+    customConfig:
+      plugins:
       - type: body-field-to-header
         name: model-extractor
-        json:
+        parameters:
           fieldName: model
           headerName: X-Gateway-Model-Name
       - type: model-provider-resolver
         name: model-provider-resolver
```
nirrozenbaum: nit: we don't need name in cases the name is identical to type (that's is set by default if no other name was specified)

deploy/payload-processing/values.yaml
```diff
@@ -10,33 +10,31 @@ upstreamBbr:
...
   provider:
-    name: istio
-    istio:
-      envoyFilter:
-        operation: INSERT_AFTER
+    name: none
 
   inferenceGateway:
-    name: maas-default-gateway
+    name: inference-gateway
```
nirrozenbaum: revert pls. this is wrong

nirrozenbaum:
```suggestion
    name: maas-default-gateway
```

deploy/payload-processing/values.yaml
```diff
@@ -10,33 +10,31 @@ upstreamBbr:
...
+          - pluginRef: apikey-injection
+          response:
+          - pluginRef: api-translation
 
   provider:
-    name: istio
-    istio:
-      envoyFilter:
-        operation: INSERT_AFTER
+    name: none
```
nirrozenbaum:
```suggestion
    name: istio
```

nirrozenbaum (approved):
/lgtm

PR #347, opened 2026-06-23
feat: add path override for cross-cluster routing

asaadbalum:
## Summary

Adds a **required** `path` field to `ExternalProviderRef` (within `ExternalModel`) that sets the outgoing request `:path` pseudo-header, enabling cross-cluster routing to a remote MaaS gateway.

- Adds required `path` field (`string`) to `ExternalProviderRef` with `MinLength=1`, `MaxLength=512`, pattern `^/.*` validation
- Path supports `{key}` placeholders resolved from the merged provider/model `Config` map (e.g. `/v1/{project}/{location}/chat/completions`)
- Propagates resolved path through `CycleState` → `PathKey`
- `api-translation` plugin reads `PathKey` and sets `:path` header (both translate and passthrough branches)

## Design

```
ExternalModel.spec.externalProviderRefs[].path (required)
  ↓ placeholders resolved from merged Config map
  ↓ stored by
model-provider-resolver → CycleState["path"]
  ↓ consumed by
api-translation → request.SetHeader(":path", path)
```

Path is required on ExternalModel (not on ExternalProvider) to align with the architectural direction where API translation is based on input/output format rather than provider type — making provider-level path assumptions invalid.

## Test Plan

- [x] Unit tests: model reconciler resolves path (placeholder resolution), CycleState propagation, header override in api-translation (translate + passthrough branches)
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] All existing tests pass (no regressions)
- [x] CRD validation: missing path rejected by API server (`Required value`)
- [x] E2E: Kind cluster + Istio + BBR deployed with full plugin chain, CRDs accepted

Closes #342

## Summary by CodeRabbit

* **New Features**
  * You can now configure a request path for external model bindings, including support for provider-specific paths and placeholder values.
  * Sample manifests were updated to show path-based routing for common providers.

* **Bug Fixes**
  * Requests now use the configured path consistently during model translation and passthrough flows.
  * Legacy migrations now preserve or infer appropriate default paths for existing provider setups.

### DIFF
--- a/api/inference/v1alpha1/externalmodel_types.go
+++ b/api/inference/v1alpha1/externalmodel_types.go
@@ -78,6 +78,16 @@ type ExternalProviderRef struct {
 	// +kubebuilder:validation:MaxLength=63
 	APIFormat string `json:"apiFormat"`
 
+	// Path sets the outgoing request `:path` pseudo-header for this model-provider binding.
+	// Must be a valid absolute path (e.g. "/v1/chat/completions" or
+	// "/maas-default-gateway/v1/chat/completions").
+	// Supports {key} placeholders resolved from the merged Config map.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	// +kubebuilder:validation:Pattern=`^/.*`
+	Path string `json:"path"`
+
 	// Config holds model-specific configuration as key-value pairs.
 	// Overrides the ExternalProvider config for this model-provider binding.
 	// +optional
--- a/deploy/samples/manifests/vertex-openai.yaml
+++ b/deploy/samples/manifests/vertex-openai.yaml
@@ -33,5 +33,6 @@ spec:
         name: vertex-openai
       targetModel: gemini-2.5-pro
       apiFormat: openai-chat
+      path: /v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions
       config:
         endpoint: openapi # Vertex AI endpoint name (not network endpoint — that's on the provider)
--- a/pkg/controller/legacymigration/reconciler.go
+++ b/pkg/controller/legacymigration/reconciler.go
@@ -117,6 +117,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 					Ref:         inferencev1alpha1.NameReference{Name: req.Name},
 					TargetModel: targetModel,
 					APIFormat:   mapProviderToAPIFormat(providerName),
+					Path:        mapProviderToDefaultPath(providerName),
 				},
 			},
 		},
@@ -205,4 +206,17 @@ func mapProviderToAPIFormat(provider string) string {
 		return "messages"
 	}
 	return "openai-chat"
+}
+
+func mapProviderToDefaultPath(provider string) string {
+	switch provider {
+	case "anthropic":
+		return "/v1/messages"
+	case "azure", "azure-openai":
+		return "/openai/v1/chat/completions"
+	case "vertex-openai":
+		return "/v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions"
+	default:
+		return "/v1/chat/completions"
+	}
 }
--- a/pkg/controller/legacymigration/reconciler_test.go
+++ b/pkg/controller/legacymigration/reconciler_test.go
@@ -318,3 +318,24 @@ func TestMapProviderToAPIFormat(t *testing.T) {
 		})
 	}
 }
+
+func TestMapProviderToDefaultPath(t *testing.T) {
+	tests := []struct {
+		provider string
+		path     string
+	}{
+		{"openai", "/v1/chat/completions"},
+		{"anthropic", "/v1/messages"},
+		{"azure", "/openai/v1/chat/completions"},
+		{"azure-openai", "/openai/v1/chat/completions"},
+		{"vertex-openai", "/v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions"},
+		{"bedrock", "/v1/chat/completions"},
+		{"bedrock-openai", "/v1/chat/completions"},
+		{"unknown", "/v1/chat/completions"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.provider, func(t *testing.T) {
+			assert.Equal(t, tt.path, mapProviderToDefaultPath(tt.provider))
+		})
+	}
+}
--- a/pkg/plugins/api-translation/plugin.go
+++ b/pkg/plugins/api-translation/plugin.go
@@ -134,6 +134,14 @@ func (p *APITranslationPlugin) WithName(name string) *APITranslationPlugin {
 	return p
 }
 
+// applyPathOverride sets the :path pseudo-header from CycleState.
+// Path is a required field on ExternalProviderRef (MinLength=1), so it is always present.
+func applyPathOverride(cycleState *plugin.CycleState, request *requesthandling.InferenceRequest) {
+	if path, err := plugin.ReadCycleStateKey[string](cycleState, state.PathKey); err == nil {
+		request.SetHeader(":path", path)
+	}
+}
+
 // ProcessRequest reads the provider from CycleState (set by an upstream plugin) and translates
 // the request body from OpenAI format to the provider's native format if needed.
 // When the incoming client format matches the upstream API format (passthrough mode),
@@ -150,6 +158,7 @@ func (p *APITranslationPlugin) ProcessRequest(ctx context.Context, cycleState *p
 		logger.Info("passthrough mode — skipping request translation")
 		// Remove client auth header; apikey-injection plugin adds the provider credential downstream.
 		request.RemoveHeader("authorization")
+		applyPathOverride(cycleState, request)
 		return nil
 	}
 
@@ -183,6 +192,8 @@ func (p *APITranslationPlugin) ProcessRequest(ctx context.Context, cycleState *p
 	// authorization is a special header removed by the plugin, no matter which provider is used.
 	// The api-key is expected to be set by the the api-key injection plugin.
 	request.RemoveHeader("authorization")
+
+	applyPathOverride(cycleState, request)
 
 	// content-length is another special header that will be set automatically by the pluggable framework when the body is mutated.
 
--- a/pkg/plugins/api-translation/plugin_test.go
+++ b/pkg/plugins/api-translation/plugin_test.go
@@ -383,6 +383,28 @@ func TestFactory_Success(t *testing.T) {
 	assert.Equal(t, APITranslationPluginType, p.TypedName().Type)
 }
 
+func TestProcessRequest_OpenAIWithPathOverride(t *testing.T) {
+	p := newTestPlugin()
+
+	cs := newCycleStateWithProvider("openai")
+	cs.Write(state.PathKey, "/maas-default-gateway/v1/chat/completions")
+
+	req := requesthandling.NewInferenceRequest()
+	req.Body["model"] = "llama-4-scout"
+	req.Body["messages"] = []any{map[string]any{"role": "user", "content": "Hi"}}
+	req.Headers["authorization"] = "Bearer sk-test"
+
+	err := p.ProcessRequest(context.Background(), cs, req)
+	require.NoError(t, err, "openai provider with path override must be supported without error")
+
+	mutated := req.MutatedHeaders()
+	assert.Equal(t, "/maas-default-gateway/v1/chat/completions", mutated[":path"],
+		"openai with path override should use the override path")
+
+	removed := req.RemovedHeaders()
+	assert.Contains(t, removed, "authorization")
+}
+
 func TestIsPassthrough(t *testing.T) {
 	tests := []struct {
 		name     string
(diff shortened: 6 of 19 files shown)


api/inference/v1alpha1/externalmodel_types.go
```diff
@@ -78,6 +78,13 @@ type ExternalProviderRef struct {
 	// +kubebuilder:validation:MaxLength=63
 	APIFormat string `json:"apiFormat"`
 
+	// Path overrides the ExternalProvider's path for this model-provider binding.
+	// When set, takes precedence over the provider-level path.
+	// +optional
+	// +kubebuilder:validation:MaxLength=512
+	// +kubebuilder:validation:Pattern=`^/.*`
+	Path string `json:"path,omitempty"`
```
nirrozenbaum: should be *string to emphasize optional field

api/inference/v1alpha1/externalmodel_types.go
```diff
@@ -78,6 +78,13 @@ type ExternalProviderRef struct {
 	// +kubebuilder:validation:MaxLength=63
 	APIFormat string `json:"apiFormat"`
 
+	// Path overrides the ExternalProvider's path for this model-provider binding.
+	// When set, takes precedence over the provider-level path.
+	// +optional
+	// +kubebuilder:validation:MaxLength=512
```
nirrozenbaum, line 87: MinLength = 1

api/inference/v1alpha1/externalprovider_types.go
```diff
@@ -58,6 +58,15 @@ type ExternalProviderSpec struct {
...
 	Auth AuthConfig `json:"auth"`
 
+	// Path overrides the outgoing request path for this provider.
+	// When set, the `:path` pseudo-header is set to this value instead of
+	// the translator's default. Must be a valid absolute path.
+	// e.g. "/maas-default-gateway/v1/chat/completions".
+	// +optional
+	// +kubebuilder:validation:MaxLength=512
+	// +kubebuilder:validation:Pattern=`^/.*`
+	Path string `json:"path,omitempty"`
```
nirrozenbaum: same comments - MinLength, pointer to string

nirrozenbaum: a general comment about path - it should be possible to put in the path placeholders for config vars. some providers require to put in the path location/project_id or even the model itself.

pkg/plugins/api-translation/plugin.go
```diff
@@ -87,6 +87,7 @@ func NewAPITranslationPlugin(ctx context.Context, config apiTranslationConfig) (
 		provider.Anthropic:     anthropic.NewAnthropicTranslator(),
 		provider.AzureOpenAI:   azure.NewAzureOpenAITranslator(),
 		provider.BedrockOpenAI: bedrock.NewBedrockOpenAITranslator(),
+		provider.RemoteMaaS:    openai.NewOpenAITranslator(),
```
nirrozenbaum: we don't need the new type RemoteMaaS. we can use openai with custom path. 
btw, as a side note - we're are moving away from mapping provider to api translator. instead we're going to map input format + output format to the right translator. so there will be no use of the provider other than display name to show clients in the UX.

asaadbalum: The remote-maas constant was only a UX label mapping to the same OpenAI translator.
openai + the new path field achieves identical behavior with no new type needed.
Updated in the latest push.

yossiovadia (commented):
Looks good overall. One minor note: the path override block is duplicated in ProcessRequest — once in the passthrough path (line ~153) and once in the normal translation path (line ~191). Could extract to a small helper, but it's only 3 lines so not blocking.

pkg/plugins/model-provider-resolver/external_provider_reconciler.go
```diff
@@ -67,6 +67,7 @@ func (r *externalProviderReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	r.store.addOrUpdateProvider(req.NamespacedName, &providerInfo{
 		provider:        provider.Spec.Provider,
 		endpoint:        provider.Spec.Endpoint,
+		path:            derefString(provider.Spec.Path),
```
nirrozenbaum: what is the fallback when path is empty?

pkg/plugins/model-provider-resolver/external_provider_reconciler_test.go
```diff
@@ -33,6 +33,8 @@ import (
 	"github.com/opendatahub-io/ai-gateway-payload-processing/pkg/plugins/common/auth"
 )
 
+func ptrString(s string) *string { return &s }
```
nirrozenbaum: pls remove. use `k8s.io/utils/ptr` instead.

pkg/plugins/model-provider-resolver/external_model_reconciler.go
```diff
@@ -109,12 +109,19 @@ func (r *externalModelReconciler) resolveRef(namespace string, ref *inferencev1a
 		weight = *ref.Weight
 	}
 
+	path := providerInfo.path
+	if ref.Path != nil && *ref.Path != "" {
+		path = *ref.Path
+	}
+	path = resolvePathPlaceholders(path, config)
```
nirrozenbaum: why this function is in a different place is that's the only usage?

pkg/plugins/model-provider-resolver/store.go
```diff
@@ -41,12 +43,32 @@ type resolvedProviderRef struct {
...
+		return ""
+	}
+	return *s
+}
+
+// resolvePathPlaceholders substitutes {key} placeholders in path with values
+// from the config map. Unresolved placeholders are left as-is.
+func resolvePathPlaceholders(path string, config map[string]string) string {
+	if path == "" || !strings.Contains(path, "{") {
+		return path
```
nirrozenbaum: what's the meaning of empty path?

pkg/plugins/model-provider-resolver/store.go
```diff
@@ -41,12 +43,32 @@ type resolvedProviderRef struct {
...
 	auth            auth.Auth
 	endpoint        string
+	path            string
 	secretName      string
 	secretNamespace string
 	config          map[string]string
 	weight          int
 }
 
+func derefString(s *string) string {
```
nirrozenbaum: where is this function used? why is it in store.go file if it's not used here?

nirrozenbaum (commented):
It's not well defined what happens when both provider and model path vars are empty. 
from the code, looks like it's doing passthrough not changing the relative path. 

we need to better document the behavior of path variable

asaadbalum:
> It's not well defined what happens when both provider and model path vars are empty. from the code, looks like it's doing passthrough not changing the relative path.
> 
> we need to better document the behavior of path variable

Yes, when both are empty, no ':path' override is applied and the original translator path is preserved.
Added documentation field comments on internal structs + updated the CRD godoc to explicitly state the "when omitted" behavior.

Also addressed the other review comments as well.

deploy/samples/manifests/vertex-openai.yaml
```diff
@@ -33,5 +33,6 @@ spec:
         name: vertex-openai
       targetModel: gemini-2.5-pro
       apiFormat: openai-chat
+      path: /v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions
```
nirrozenbaum, line 36: nice

pkg/controller/legacymigration/reconciler.go
```diff
@@ -206,3 +207,14 @@ func mapProviderToAPIFormat(provider string) string {
...
 	return "openai-chat"
 }
+
+func mapProviderToDefaultPath(provider string) string {
+	switch provider {
+	case "anthropic":
+		return "/v1/messages"
+	case "azure", "azure-openai":
+		return "/openai/v1/chat/completions"
+	default:
```
nirrozenbaum, line 219: can you add also vertex?
(the same format you had on the sample)

pkg/controller/legacymigration/reconciler_test.go
```diff
@@ -318,3 +318,24 @@ func TestMapProviderToAPIFormat(t *testing.T) {
...
+func TestMapProviderToDefaultPath(t *testing.T) {
+	tests := []struct {
+		provider string
+		path     string
+	}{
+		{"openai", "/v1/chat/completions"},
+		{"anthropic", "/v1/messages"},
+		{"azure", "/openai/v1/chat/completions"},
+		{"azure-openai", "/openai/v1/chat/completions"},
+		{"vertex-openai", "/v1/chat/completions"},
```
nirrozenbaum: this is incorrect

pkg/plugins/model-provider-resolver/plugin_test.go
```diff
@@ -93,6 +93,78 @@ func TestProcessRequest_ModelResolved(t *testing.T) {
...
+	store := newInfoStore()
+	store.addOrUpdateModel(
+		types.NamespacedName{Namespace: "llm", Name: "gpt4"},
+		&externalModelInfo{modelName: "gpt4", refs: []*resolvedProviderRef{{
+			provider:        provider.OpenAI,
+			targetModel:     "gpt-4o",
+			apiFormat:       apiformat.OpenAIChatCompletions,
+			auth:            auth.Simple,
+			endpoint:        "api.openai.com",
+			path:            "",
```
nirrozenbaum: what is this test checking? 
path cannot be empty.. it's a required field on the CRD with MinLength=1

test/e2e/e2e_test.go
```diff
@@ -85,7 +85,21 @@ spec:
...
+}
+
+func defaultPathForProvider(provider string) string {
+	switch provider {
+	case "anthropic":
+		return "/v1/messages"
+	case "azure-openai":
+		return "/openai/v1/chat/completions"
+	case "vertex-openai":
+		return "/v1/projects/test-project/locations/us-central1/endpoints/openapi/chat/completions"
```
nirrozenbaum, line 99: here it was updated correctly :)

pkg/plugins/api-translation/plugin.go
```diff
@@ -134,6 +134,14 @@ func (p *APITranslationPlugin) WithName(name string) *APITranslationPlugin {
 	return p
 }
 
+// applyPathOverride sets the :path pseudo-header from CycleState.
+// Path is a required field on ExternalProviderRef, so it is always present.
+func applyPathOverride(cycleState *plugin.CycleState, request *requesthandling.InferenceRequest) {
+	if path, err := plugin.ReadCycleStateKey[string](cycleState, state.PathKey); err == nil && path != "" {
```
nirrozenbaum: nit: path is required and with MinLength=1.. so it cannot be ever empty:
 
```suggestion
	if path, err := plugin.ReadCycleStateKey[string](cycleState, state.PathKey); err == nil {
```

pkg/plugins/api-translation/plugin_test.go
```diff
@@ -413,6 +435,75 @@ func TestIsPassthrough(t *testing.T) {
...
+}
+
+func TestProcessRequest_NoPathOverride_KeepsTranslatorDefault(t *testing.T) {
+	mock := &mockTranslator{
+		reqHeaders: map[string]string{":path": "/v1/chat/completions"},
+	}
+	p := newPluginWithMock("openai", mock)
+
+	cs := newCycleStateWithProvider("openai")
+	cs.Write(state.PathKey, "")
```
nirrozenbaum: can path be ever empty?

pkg/plugins/model-provider-resolver/store.go
```diff
@@ -41,12 +43,32 @@ type resolvedProviderRef struct {
...
+// resolvePathPlaceholders substitutes {key} placeholders in path with values
+// from the config map. Unresolved placeholders are left as-is.
+func resolvePathPlaceholders(path string, config map[string]string) string {
+	if path == "" || !strings.Contains(path, "{") {
+		return path
+	}
+	for k, v := range config {
+		path = strings.ReplaceAll(path, "{"+k+"}", v)
+	}
+	return path
```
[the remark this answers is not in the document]
nirrozenbaum: @asaadbalum a point to pay attention to. 
you can defer to follow up not to stall this PR (but please open an issue so we track it)

noyitz:
Addressed all review comments in 66fe0ea:

- **Path → `*string` (optional)** on both `ExternalProviderRef` and `ExternalProviderSpec`, with `+optional` and `MinLength=1` markers. When omitted, no `:path` override is applied.
- **Added `vertex-openai`** case to `mapProviderToDefaultPath` with the correct path template (`/v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions`).
- **Fixed vertex-openai test** — was expecting `/v1/chat/completions`, now expects the correct vertex path.
- **Updated deepcopy** (`zz_generated.deepcopy.go`) for `*string` Path fields on both structs.
- **Renamed empty-path test** to `NoPathOverrideWrittenToCycleState` with clarifying comments (empty string = Path was omitted, no override).
- **Updated `applyPathOverride` comment** to document the empty path behavior.
- **All `Path:` assignments** use `ptr.To()` consistently across tests.

Build and vet pass clean. `make generate manifests` still needs to be run to regenerate CRD YAMLs.

noyitz:
Reverted my commit (66fe0ea) — it unintentionally undid some of the existing fixes. Apologies for the noise. Leaving this for Asaad and Senan to close out.

nirrozenbaum (approved):
/lgtm
/approve

Thanks!

PR #349, opened 2026-06-23
feat: add routing loop detection for cross-cluster requests

### LINKED ISSUE #343: feat: add routing loop detection for cross-cluster requests
## Parent Issue

Part of #276 (Cross-cluster inference routing via ExternalModel/Provider)

Depends on #342 (path field and remote-maas provider)

## Summary

Add routing loop detection to prevent infinite request cycles when Cluster A routes to Cluster B and Cluster B (by misconfiguration) routes back to Cluster A.

### Logic added to model-provider-resolver plugin

**Incoming request check (early in ProcessRequest):**
- Check for `X-Origin-Cluster` header on incoming requests
- If present and value matches this cluster's identity, reject with HTTP 508 Loop Detected
- If present but different cluster, allow (normal cross-cluster request)
- If absent, allow (normal request)

**Outgoing cross-cluster injection (after resolving to a remote-maas provider):**
- When CycleState has `provider: remote-maas`, inject `X-Origin-Cluster: <this-cluster-name>` header into the outgoing request

### Cluster identity
- Sourced from environment variable (`CLUSTER_NAME`) on the IPP pod
- Configurable via Helm values / deployment manifest

## Testing

- Request with no `X-Origin-Cluster` header: passes through
- Request with `X-Origin-Cluster: other-cluster`: passes through
- Request with `X-Origin-Cluster: this-cluster`: rejected with 508
- Outgoing cross-cluster request gets `X-Origin-Cluster` header injected
- Non-cross-cluster requests (local models) are unaffected

## Acceptance Criteria

- A → B → A loop is detected and rejected with HTTP 508 Loop Detected
- Normal cross-cluster requests (A → B, no loop) work without issue
- Non-cross-cluster requests (local models) are completely unaffected
- Cluster identity is configurable via environment variable

abdallahsamabd:
Adds routing loop detection to the model-provider-resolver plugin to prevent infinite request cycles in cross-cluster routing scenarios.

Simple rule: a request that has been forwarded once cannot be forwarded again.

The `X-Gateway-Forwarded: true` header is set on every ExternalModel resolution. If a subsequent payload processor sees this header on an incoming request that also resolves to an ExternalModel, the request is rejected — it has already been forwarded once and must be served locally (via InferenceService), not through another ExternalModel.

**Loop check (after model/provider resolution in ProcessRequest):**

- If `X-Gateway-Forwarded: true` header is present AND the request resolves to an ExternalModel → rejects with HTTP 403 (request already forwarded, cannot be processed through another ExternalModel)
- If `X-Gateway-Forwarded: true` header is present AND the request does NOT match any ExternalModel → passes through (served locally via InferenceService — no loop)
- If `X-Gateway-Forwarded` header is absent → allows (normal request, first hop)

**Outgoing injection (after ExternalModel resolution):**

- Sets `X-Gateway-Forwarded: true` on every ExternalModel resolution
- Header is set once — harmless for non-MaaS endpoints (external APIs like OpenAI don't run our payload processor)

**No configuration required:** loop detection is always on. No env vars, no cluster identity resolution — just a simple boolean header.

**Covers loop scenarios:**

- **Self-loops (A → A):** A sets header, receives its own request back, resolves to ExternalModel → blocked
- **Two-cluster loops (A → B → A):** A sets header, B receives request with header, if B resolves to ExternalModel → blocked
- **Any multi-hop attempt (A → B → C):** B receives request with header, if B tries to resolve another ExternalModel → blocked (there is no legitimate use-case for multi-hop cross-cluster routing; A can go directly to C)

Depends on #347
Closes #343

## How Has This Been Tested?

5 new unit tests covering all loop detection scenarios:

- **TestProcessRequest_ForwardedRequest_Blocked** — request with `X-Gateway-Forwarded: true` + ExternalModel resolution → 403 error
- **TestProcessRequest_ForwardedRequest_BeforeCycleStateWrite** — loop rejected before CycleState is written (no side effects)
- **TestProcessRequest_ForwardedRequest_InternalModel_PassesThrough** — forwarded header present but model is NOT in ExternalModel store → passes through (served locally via InferenceService)
- **TestProcessRequest_NoForwardedHeader_Passes** — no header → passes through normally
- **TestProcessRequest_ExternalModel_InjectsForwardedHeader** — first ExternalModel resolution → forwarded header injected

All 50 tests pass (5 new + 45 existing, 0 regressions).

### Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **Bug Fixes**
  * Added protection against routing loops for requests that are forwarded more than once.
  * Requests sent through external model routing now return a clear forbidden error if a loop is detected.
  * Outgoing requests now include a forwarded marker so repeated passes can be identified.
* **Tests**
  * Added coverage for forwarded-request handling, loop prevention, and normal external/internal routing behavior.


noyitz:
Not targeting 3.5 — cross-cluster routing is an AI Grid / 3.6 feature. Candidate for 3.6.

szedan-rh:
Thanks Noy, Making sense, let's target 3.6

abdallahsamabd:
fixed:
Added x-gateway-forwarded to maas-headers-guard: we capture it into CycleState then strip it from the request (same pattern as x-maas-*). A bare strip would drop the hop-2 marker before model-provider-resolver runs, since the guard is first in the profile. Resolver now reads the captured value from CycleState.
regarding third comment,
the loop-detection test now type-asserts errcommon.Error and checks .Code == errcommon.Forbidden

noyitz (commented):
Reviewed the loop-detection logic in depth, with focus on spoofing. The security posture is the strong part and passes; the concerns are design alignment with issue #343 and collision with the in-flight hub-mode PRs.

### Security (spoofing) — passes
Traced the full chain. `maas-headers-guard` runs before `model-provider-resolver` (`deploy/payload-processing/values.yaml:16-36`), captures `x-gateway-forwarded` into CycleState **then** strips it from the upstream request before the resolver runs — so a client-supplied value never leaks upstream but is preserved for the check. A client *can* set `X-Gateway-Forwarded: true`, but the guard is fail-closed: presence only ever causes a 403, never a bypass, and inducing a real loop would require *clearing* the marker at each hop (gateway-controlled, not client-controlled). Self-DoS on one's own request is the only downside — worth a one-line doc note. `Forbidden → 403` maps correctly through `errcommon`.

### Blocking

**1. Divergence from linked issue #343 wrongly rejects legitimate multi-hop chains.** #343 specifies per-cluster identity — inject `X-Origin-Cluster: <CLUSTER_NAME>` and reject (508) only when origin == *this* cluster (a true A→B→A loop-back). This PR implements a boolean "forwarded once → never forward to an external model again." Consequence: a legitimate **A→B→C** chain (B forwards to a *different* cluster C, no loop) is 403'd, because B sees `x-gateway-forwarded=true` and C resolves to an ExternalModel. The PR description's "no legitimate multi-hop use-case" contradicts the linked issue's acceptance criteria — this needs an explicit team decision (forward-once vs. per-cluster-identity), not a silent redefinition. CodeRabbit's linked-issue check flagged the same.

**2. Collision with #415/#416 (hub mode), textual and semantic.** All three edit `model-provider-resolver/plugin.go` `ProcessRequest` + the same import block. #415 restructures `ProcessRequest` into PROPOSE/TRANSFORM; this inserts `checkRoutingLoop`/`markForwarded` into the same function — they will textually conflict. Semantically, hub→spoke forwarding is intrinsic: in TRANSFORM the hub resolves to a spoke and would `markForwarded`, and the spoke re-resolving to an external provider would then 403. "Forward once" is fundamentally at odds with the hub model. Whichever merges second must reconcile the loop logic against the phase model (e.g. skip loop detection in hub PROPOSE/TRANSFORM), not just fix the textual conflict.

### Non-blocking
- 508 isn't expressible in `errcommon` (no `LoopDetected`); 403 is a reasonable fallback and the code comments acknowledge it. If 508 is required by the spec, `errcommon` must be extended upstream.
- `markForwarded` sets `x-gateway-forwarded` on *every* external resolution, including direct calls to third-party APIs (e.g. `api.openai.com`) — harmless but leaks an internal header externally. Consider injecting only for remote-maas/cluster targets (also aligns with #343's "when provider is remote-maas").
- Header value compare is case-sensitive (`"true"`) while the key is case-insensitive — a spoofed `TRUE` wouldn't match. No security impact (producer emits lowercase, and the miss is in the allow direction), but worth normalizing.
- The header name literal is defined in three places (state key, guard, resolver), kept in sync by comments only.
- Tests cover each branch well; gaps: end-to-end guard+resolver spoofing test, an A→B→C false-positive test (would document the #343 divergence), and a value-case test.

Given it's already on `do-not-merge/hold` for 3.6, the forward-once-vs-#343 decision is the thing to settle before this comes off hold.

szedan-rh (approved):
LGTM — both review comments are addressed in the latest revision:

1. **Untrusted-header access decision (CWE-807):** loop detection no longer trusts a raw client header. It now keys off cluster identity (`CLUSTER_NAME`) and is disabled when that is unset. `maas-headers-guard` captures `x-origin-cluster` into CycleState and strips it from the request before it can reach any upstream; the resolver reads the captured value from CycleState. A→B→A is rejected, A→B→C is allowed.

2. **Test status-code assertion:** the loop-detection tests now go through `requireLoopForbidden`, which type-asserts `errcommon.Error` and checks `.Code == errcommon.Forbidden` rather than only matching the message substring. 403/Forbidden is used deliberately (documented in code) since `errcommon` has no 508 code.

`lint-and-test` is green. Nice work.

PR #354, opened 2026-06-25
fix: wire up ExternalModel HTTPRoute reconciler with gateway config from plugin JSON

ishitasequeira:
## Description
- Parse gatewayName/gatewayNamespace from the model-provider-resolver plugin config JSON (previously ignored via _)
- Wire up the ExternalModel HTTPRoute reconciler and ExternalProvider networking reconciler alongside the existing store-based reconcilers in the plugin factory
- Add gateway config to production Helm values so HTTPRoutes reference `maas-default-gateway` instead of the hardcoded default default-gateway       
   
## Problem                                                                     
The model-provider-resolver plugin discarded its config JSON and never started the resource-creating reconcilers. This meant:
1. No HTTPRoutes were created for `inference.opendatahub.io/ExternalModel` CRs in production
2. Even when the reconciler ran (e.g. via test setups), HTTPRoutes got parentRef: `default-gateway` instead of `maas-default-gateway`, causing `MaaSSubscription` to reject them with: HTTPRoute does not reference tenant Gateway `openshift-ingress/maas-default-gateway`

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **Chores**
  * Updated deployment settings for the payload-processing service to use a specific gateway configuration for model routing.

### DIFF
--- a/deploy/payload-processing/values.yaml
+++ b/deploy/payload-processing/values.yaml
@@ -21,6 +21,10 @@ upstreamIpp:
           fieldName: model
           headerName: X-Gateway-Model-Name
       - type: model-provider-resolver
+        name: model-provider-resolver
+        parameters:
+          gatewayName: "maas-default-gateway"
+          gatewayNamespace: "openshift-ingress"
       - type: api-translation
       - type: apikey-injection
       profiles:
--- a/deploy/payload-processing/values.yaml
+++ b/deploy/payload-processing/values.yaml
@@ -21,7 +21,6 @@ upstreamIpp:
           fieldName: model
           headerName: X-Gateway-Model-Name
       - type: model-provider-resolver
-        name: model-provider-resolver
         parameters:
           gatewayName: "maas-default-gateway"
           gatewayNamespace: "openshift-ingress"
--- a/pkg/plugins/model-provider-resolver/plugin.go
+++ b/pkg/plugins/model-provider-resolver/plugin.go
@@ -34,8 +34,11 @@ import (
 	errcommon "github.com/llm-d/llm-d-inference-payload-processor/pkg/common/error"
 	logutil "github.com/llm-d/llm-d-inference-payload-processor/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	inferencev1alpha1 "github.com/opendatahub-io/ai-gateway-payload-processing/api/inference/v1alpha1"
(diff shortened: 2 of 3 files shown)


deploy/payload-processing/values.yaml
```diff
@@ -21,6 +21,10 @@ upstreamIpp:
           fieldName: model
           headerName: X-Gateway-Model-Name
       - type: model-provider-resolver
+        name: model-provider-resolver
```
nirrozenbaum: no need for `name` when it's identical to `type` (config loader sets it by default when omitted)

nirrozenbaum:
@ishitasequeira we have separate controllers for setting the resources for clear separation of concerns. 
controllers who watch ExternalProvider/Model are control plane controllers - those are initialized here:
https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/be0f0085f4b58b81bd56531bc3a02c1416f58c1a/cmd/main.go#L40-L46
these are responsible for creating the networking resources (ServiceEntry, DestinationRule, HttpRoute, ..)

the place where you've added controllers is the plugins that run in data plane.. 
we do hook there controllers, but ones which collect information that is used by the plugin (e.g., reconciler that gets information from the CRs and saves it inside in-memory cache.

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -47,17 +47,18 @@ const (
 var _ requesthandling.RequestProcessor = &ModelProviderResolverPlugin{}
 
 // ModelProviderResolverFactory defines the factory function for ModelProviderResolverPlugin.
-func ModelProviderResolverFactory(name string, _ json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
-	plugin, err := NewModelProviderResolver(handle.ReconcilerBuilder, handle.Client())
+func ModelProviderResolverFactory(name string, rawConfig json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
```
nirrozenbaum: can we revert the changes in this file as well? 
rawConfig is not used so we can keep the wildcard. 

I think the part you wanted to fix is here:
https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/be0f0085f4b58b81bd56531bc3a02c1416f58c1a/pkg/controller/common/constants.go#L25

ishitasequeira:
Closing in favor of https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/356/changes#diff-4eb73b70655bc28654d85407ce0b78de240d194270de3d54c74be7d9a21b676bR25

PR #359, opened 2026-06-25
feat: resolve ExternalModel by body model name for single-URL routing

noyitz:
## Summary
- Replace path-based resolution (extracting namespace/name from URL segments) with body-model-name lookup via `getModelByName`
- Enables single-URL routing (`POST /v1/chat/completions` with model only in body) — no model in the URL path
- Resolver prefers `X-Gateway-Model-Name` header (set by body-field-to-header) with fallback to body `model` field
- Fixes `detectInputAPIFormat` for bare paths after `sanitizePath` strips the leading slash

## Context
MaaS already supports single-URL routing:
- PR #948 (MaaS) adds pre-auth ext_proc that sets `X-Gateway-Model-Name` from the body
- MaaS controller generates header-based HTTPRoute rules matching on `X-Gateway-Model-Name`

The only gap was in the IPP resolver which only did path-based extraction. This PR closes that gap.

## Test plan
- [ ] Single URL `/v1/chat/completions` with `model: gpt-5.5` resolves to correct ExternalModel
- [ ] Single URL `/v1/messages` with `model: claude-haiku-4-5-20251001` resolves correctly
- [ ] Per-model URLs (`/llm/ext-openai/v1/chat/completions`) continue to work via body model fallback
- [ ] Models with same name in different namespaces resolve correctly

## Risk analysis
**Risk rating:** 2
**Why:** Changes resolution logic in model-provider-resolver which is on the critical path. However, the change is straightforward (lookup by modelName field instead of path segments) and both URL patterns are covered.

## Summary by CodeRabbit

* **New Features**
  * Model selection now resolves by model name from the request header, with a fallback to the request body when the header is not provided.
* **Bug Fixes**
  * Improved model lookup to find matching models by name across namespaces.
  * Relaxed strict model name validation to avoid rejecting requests when naming doesn’t match exactly.
  * Updated the “unsupported API” error message to be clearer and no longer include the raw endpoint path.

### DIFF
--- a/pkg/plugins/model-provider-resolver/plugin.go
+++ b/pkg/plugins/model-provider-resolver/plugin.go
@@ -136,30 +136,24 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 	log.FromContext(ctx).V(logutil.VERBOSE).Info("received incoming request", "path", request.Headers[":path"])
 	relativePath := sanitizePath(request.Headers[":path"])
 
-	segments := strings.Split(relativePath, "/")
-	if len(segments) < 2 || segments[0] == "" || segments[1] == "" {
-		log.FromContext(ctx).V(logutil.VERBOSE).Info("wasn't able to parse namespaced name from path", "path", relativePath)
-		return nil
+	// Resolve by model name: prefer X-Gateway-Model-Name header (set by body-field-to-header),
+	// fall back to request body model field. This supports both single-URL and per-model-URL patterns.
+	modelName := request.Headers["x-gateway-model-name"]
+	if modelName == "" {
+		modelName = model
 	}
 
-	modelKey := types.NamespacedName{Namespace: segments[0], Name: segments[1]}
-	log.FromContext(ctx).V(logutil.VERBOSE).Info("exported namespaced name from path", "key", modelKey)
-
-	modelInfo, found := p.store.getModel(modelKey)
+	modelInfo, modelKey, found := p.store.getModelByName(modelName)
 	if !found {
 		return nil // not an external model — pass through for internal models
 	}
 
+	logger.Info("resolved model by name", "modelName", modelName, "key", modelKey.String())
+
 	inputFormat := detectInputAPIFormat(relativePath)
 	if inputFormat == "" {
 		logger.Error(nil, "unsupported API path for external model", "model", modelKey.String(), "path", relativePath)
-		return errcommon.Error{Code: errcommon.BadRequest, Msg: fmt.Sprintf("unsupported API path: %s", relativePath)}
-	}
-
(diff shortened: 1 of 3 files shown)


nirrozenbaum:
please rebase

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -182,11 +176,11 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
 // detectInputAPIFormat determines the client's API format from the request path suffix.
 func detectInputAPIFormat(path string) apiformat.APIFormat {
 	switch {
-	case strings.HasSuffix(path, "/v1/chat/completions"):
+	case strings.HasSuffix(path, "/v1/chat/completions"), path == "v1/chat/completions":
```
nirrozenbaum: why do you add path == ... if it's already covered in the `HasSuffix`?

pkg/plugins/model-provider-resolver/store.go
```diff
@@ -129,3 +129,18 @@ func (s *infoStore) getModel(key types.NamespacedName) (*externalModelInfo, bool
 	info, ok := modelsByNamespace[key.Name]
 	return info, ok
 }
+
+// getModelByName finds an ExternalModel by its client-facing modelName across all namespaces.
+func (s *infoStore) getModelByName(modelName string) (*externalModelInfo, types.NamespacedName, bool) {
```
nirrozenbaum: this is a wrong solution. 
instead we should flatten the store to completely remove ns from it. ns is no longer relevant as models are identified by their unique name.

pkg/plugins/model-provider-resolver/plugin.go
```diff
@@ -136,30 +136,24 @@ func (p *ModelProviderResolverPlugin) ProcessRequest(ctx context.Context, cycleS
...
 
-	segments := strings.Split(relativePath, "/")
-	if len(segments) < 2 || segments[0] == "" || segments[1] == "" {
-		log.FromContext(ctx).V(logutil.VERBOSE).Info("wasn't able to parse namespaced name from path", "path", relativePath)
-		return nil
+	// Resolve by model name: prefer X-Gateway-Model-Name header (set by body-field-to-header),
+	// fall back to request body model field. This supports both single-URL and per-model-URL patterns.
+	modelName := request.Headers["x-gateway-model-name"]
+	if modelName == "" {
+		modelName = model
```
nirrozenbaum, line 143: this is not needed. you've already read model from the body so just use it.
no point in this extra logic.

nirrozenbaum: not worth blocking the PR on this, but I suggest cleaning up in a follow up (the shorter the code is, the easier to maintain it).

PR #360, opened 2026-06-25
feat: refactor api-translation to select by (inputFormat, outputFormat)

yossiovadia:
Refactors api-translation plugin per Nir's direction: translator selection based on (inputFormat, outputFormat) tuple instead of provider name. Provider is completely removed from translation logic.

- Registry: map[translatorKey]translator where key is {inputFormat, outputFormat}
- ProcessRequest/ProcessResponse: read InputAPIFormatKey + APIFormatKey from CycleState
- Currently registered: (openai-chat, openai-chat) and (openai-chat, messages)
- Azure/Bedrock/Vertex translators to be re-added as needed with appropriate format keys

2 files, net -21 lines. All tests pass, lint clean.

Closes #316

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API translation routing so requests and responses are matched by their input/output format combination instead of provider name.
  * Added clearer errors for unsupported format pairs, making translation failures easier to understand.
  * Translation now consistently skips passthrough mode and continues to remove authorization data from outgoing requests.
  * Updated response handling to use the current model and apply translated response bodies more reliably.


pkg/plugins/api-translation/plugin.go
```diff
@@ -78,49 +73,35 @@ func APITranslationFactory(name string, rawConfig json.RawMessage, handle plugin
...
+}
 
-	keys := make([]string, 0, len(providers))
-	for key := range providers {
-		keys = append(keys, key)
+func NewAPITranslationPlugin(ctx context.Context, config apiTranslationConfig) (*APITranslationPlugin, error) {
+	translators := map[translatorKey]translator.Translator{
+		// OpenAI chat → provider-specific format
+		{apiformat.OpenAIChatCompletions, apiformat.OpenAIChatCompletions}: openai.NewOpenAITranslator(),
+		{apiformat.OpenAIChatCompletions, apiformat.Messages}:             anthropic.NewAnthropicTranslator(),
```
nirrozenbaum, line 64: the logic in this file should change. 
we should have in this map only real translations. 
e.g., ChatCompletionsToMessages. 
we need to rename the translators. no more `Anthropic`. or `OpenAI`. it should be driven by the API. 
in our case we can start with a single translator `ChatCompletionsToMessages` which can be in a go file under translator dir. 
all provider dirs should be removed. 

given a request - if input format == output format, we skip the translator. 
otherwise we can call the right one according to the mapping. 
this would give us to ability to support any passthrough (without explicitly specifying it), and have auto translation between pre defined APIs.

yossiovadia: Makes sense — I'll keep this PR as the mapping refactor (provider → format tuple), then follow up with a PR that renames the translators by API format and removes the provider directories. That way we get incremental review instead of one large diff.

